PR #6, opened 2022-10-28
feat(home-screen):home screen and tests for home screen+app added

atticusofsparta:
Merge the home screen and test's to develop


src/__tests_/App.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { render, fireEvent, cleanup } from 'test-utils';
+import '@testing-library/jest-dom';
+import App from '../App';
+
+afterEach(cleanup);
+
+test('render App', () => {
+  render(<App />);
+});
```
dtfiedler, line 10: nit: wrap all tests in relevant `describe` blocks

```
describe('App', () => {
  afterEach(cleanup);
  
  test('render App', () => {
    render(<App />);
  });
});
```

describes allow for effective scoping within test suites

src/__tests_/App.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
dtfiedler, line 1: you can add 
```
import React from 'react';
global.React = React; 
```
to `jest.config.js` to set this for all test files so we don't have to import it everytime.

src/__tests_/App.test.tsx
```diff
@@ -0,0 +1,10 @@
...
+import { render, fireEvent, cleanup } from 'test-utils';
```
dtfiedler, line 2:
```suggestion
import { render, cleanup } from 'test-utils';
```
nit: remove unused dependencies/imports - we'll eventually have a eslint rule that checks for this

src/components/pages/Home/__tests__/Home.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { render, fireEvent, cleanup } from 'test-utils';
+import '@testing-library/jest-dom';
+import Home from '../Home';
+
+afterEach(cleanup);
+
+test('render Home', () => {
+  render(<Home />);
+});
```
dtfiedler, line 10: same comment, wrap in a describe block

src/__tests_/App.test.tsx
```diff
@@ -0,0 +1,10 @@
...
+import '@testing-library/jest-dom';
```
dtfiedler, line 3: i don't think this is necessary as it's imported in `jest.config.js`

src/App.tsx
```diff
@@ -1,32 +1,15 @@
...
-      <p className="read-the-docs">
-        Click on the Vite and React logos to learn more
-      </p>
+    <div>
+      <Routes>
+        <Route path="*" element={<Home />} />
+        <Route path="/Home" element={<Home />} />
+      </Routes>
     </div>
   );
```
dtfiedler, line 14: i think we can avoid some of the testing wrappers by doing it like this:

```
import { Routes, Route, BrowserRouter as Router } from 'react-router-dom';
import './index.css';
import Home from './components/pages/Home/Home';

function App() {
  return (
      <Router>
        <Routes>
          <Route path="*" element={<Home />} />
          <Route path="/Home" element={<Home />} />
        </Routes>
      </Router>
    )
}
````

File: src/utils/test-utils.tsx [hunk 7]
```diff
@@ -0,0 +1,15 @@
...
+  return <BrowserRouter>{children}</BrowserRouter>;
+};
+
+const customRender = (
+  ui: ReactElement,
+  options?: Omit<RenderOptions, 'wrapper'>,
+) => render(ui, { wrapper: AllTheProviders, ...options });
+
+export * from '@testing-library/react';
+export { customRender as render };
```
dtfiedler, line 15: i understand why this is here, but I think you can eliminate this with the use of `BrowserRouter` in App.jsx instead of just `Routes`. See above suggestion for that.

src/__tests_/App.test.tsx
dtfiedler, line 2: with the use of `BrowserRouter` we no longer need to use the wrappers, so we can reference the imports directly from the testing-library

```suggestion
import { render, cleanup } from '@testing-library/react';
```

src/components/pages/Home/__tests__/Home.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { render, fireEvent, cleanup } from 'test-utils';
```
dtfiedler, line 2:
```suggestion
import { render, cleanup } from '@testing-library/react';
```

src/components/pages/Home/__tests__/Home.test.tsx
```diff
@@ -0,0 +1,10 @@
...
+import '@testing-library/jest-dom';
```
dtfiedler, line 3:
```suggestion
```

dtfiedler (changes requested):
some small changes r.e. testing configuration and setup - but the app is loading and the components themselves look good. feel free to follow up with any unclear suggestions

src/__tests_/App.test.tsx
atticusofsparta: Correct, works without it

src/App.tsx
atticusofsparta: correct, works without it. Leaving an example in a comment in test-utils.tsx though.

src/__tests_/App.test.tsx
atticusofsparta: Also correct, changed

src/__tests_/App.test.tsx
atticusofsparta: all current comment changes added in commit 17f4d4f5850e37a51176f3566e276723b217dc34

dtfiedler (approved):
looks good. Building the base, and iterating!

PR #10, opened 2022-10-31
feat(gh-actions): build and test arns app

atticusofsparta:
actions setup to build and test on pushes to any branch


.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,14 @@
+name: Build and Test
+
+on:
+  push:
+    branches: ['**']
```
dtfiedler, line 5: not needed - you remove this whole line and just have it as `push`

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  test_build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
dtfiedler, line 11: Let's use v3

dist/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
```
dtfiedler, line 2: Please add `dist` to .gitignore

dtfiedler (changes requested):

.github/workflows/build_and_test.yml
atticusofsparta: add all requested changes in commit 0b1d5327776ed08bb564010ec564045a84dfde2e

dtfiedler (approved):
looks good, think we got the jest stuff sorted. we'll probably want to add prettier checks in here to keep the codebase clean. could be a separate action that we only check on merges

PR #13, opened 2022-11-01
feat(pages): add pages for about and faq

atticusofsparta:


src/components/pages/About/__tests__/About.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+describe('About', () => {
+  afterEach(cleanup);
+
+  test('render About', () => {
+    render(
+      <Router>
+        <About />
+      </Router>,
+    );
+  });
```
dtfiedler, line 14: we don't need to wrap these pages in `Router` only NavBar since it uses `Link`

src/components/pages/About/__tests__/About.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+});
```
dtfiedler, line 15: nit: we don't need to wrap these pages in `Router` only NavBar since it uses `Link`
```suggestion
import { render, cleanup } from '@testing-library/react';
import About from '../About';

describe('About', () => {
  afterEach(cleanup);

  test('render About', () => {
    render(<About />);
  });
});
```

src/App.tsx
```diff
@@ -1,6 +1,8 @@
 import { Routes, Route, BrowserRouter as Router } from 'react-router-dom';
 import './index.css';
 //pages
+import About from './components/pages/About/About';
+import FAQ from './components/pages/FAQ/FAQ';
 import Home from './components/pages/Home/Home';
 //layout
 import NavBar from './components/layout/Navbar/Navbar';
```
dtfiedler, line 8: we can create `index.ts` in `/components/pages` that exports all of our pages from there, so we are then able to import pages a bit more concisely

e.g.
```
import { About, FAQ, Home } from './components/pages'`
```

atticusofsparta: requested changes added in commit a725b9f156d13fb7fc917371c2ff64d339103f87

dtfiedler (approved):
looks good!

PR #14, opened 2022-11-02
feat(routing): update routing to use createBrowserRouter, add layouts for header and footers

dtfiedler:
### Details
- Moves us to using `createBrowserRouter` in place of `BrowserRouter`
- Adds `Layout` component which wraps every page, to include header and footer
- Imports `styles.css` to pages that were not importing them before

### DIFF
--- a/src/components/layout/Layout/Layout.tsx
+++ b/src/components/layout/Layout/Layout.tsx
@@ -1,8 +1,8 @@
 import NavBar from '../Navbar/Navbar';
 import Footer from '../Footer/Footer';
-import './styles.css'; // will be injected into the page
+import './styles.css';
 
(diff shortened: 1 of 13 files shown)


src/components/layout/Layout/Layout.tsx
```diff
@@ -0,0 +1,24 @@
...
+        <div className="container">
+            <Outlet />
+        </div>
+        <div className="footer">
+          <Footer />
+        </div>
+      </div>
+    </>
+  );
+}
```
dtfiedler, line 22: this allows us to add headers/footers/side panels to every page. children components (pages) get happened in `Outlet`

atticusofsparta (approved):
looks good

PR #18, opened 2022-11-02
feat(arns): create interface for smartweave source code contract fetc…

dtfiedler:
…hing, implement a local file system class to fetch local contract


atticusofsparta (commented):
Might be a good idea to name txid in the SmartweaveContractState type to contractId instead. Helps code comprehension without knowledge of smartweave. And we may have other interfaces that take in txid, so this could make a clear difference.

atticusofsparta (approved):

PR #21, opened 2022-11-04
feat(type): add GlobalState type

atticusofsparta:


src/types.ts
```diff
@@ -1,5 +1,12 @@
 import React from 'react';
 
+
+export type GlobalState = {
+  arnsSourceContract: ArNSContractState;
+  gateway:string;
+  connectedWallet: string;
+  errors:Array<Error>;
+}
```
dtfiedler, line 9: the type looks good - but let's move it to `src/state/state.tsx`. I'll create another ticket to port over some of the helpers/reducers similar to price-calculator app : https://github.com/ardriveapp/price-calculator/blob/master/src/state/state.tsx

dtfiedler (approved):

PR #22, opened 2022-11-04
feat(globalState): add GlobalState provider and wrap App component in it

atticusofsparta:


src/state/reducer.ts
```diff
@@ -0,0 +1,36 @@
...
+    case 'setArnsContractState':
+      return {
+        ...state,
+        arnsSourceContract: action.payload,
+      };
+    case 'setErrors':
+      return {
+        ...state,
+        errors: action.payload,
+      };
```
dtfiedler, line 31: we're going to want to be able to push errors and remove errors from the array, so for now, let's move this
```suggestion
```

src/state/state.tsx
```diff
@@ -1,8 +1,44 @@
...
 
 export type GlobalState = {
   arnsSourceContract: ArNSContractState;
   gateway: string;
   connectedWallet: string;
   errors: Array<Error>;
 };
+
+const initialState: GlobalState = {
+  arnsSourceContract: testContractState,
```
dtfiedler, line 14: let's start with 
```suggestion
  arnsSourceContract: { records: {} }
```
and remove the direct import from the local file (we want to use `LocalDataProvider` to fetch this file dynamically)

src/state/state.tsx
```diff
@@ -1,8 +1,44 @@
...
+  gateway: 'arweave.net',
+  connectedWallet: '',
```
dtfiedler, line 16:
```suggestion
  connectedWallet: undefined
```

src/state/reducer.ts
```diff
@@ -0,0 +1,36 @@
+import { GlobalState } from './state';
+import type { ArweaveTransactionId, ArNSContractState } from '../types';
+
+export type Action =
+  | { type: 'setWalletAddress'; payload: ArweaveTransactionId }
+  | { type: 'setGateway'; payload: string }
+  | { type: 'setArnsContractState'; payload: ArNSContractState }
+  | { type: 'setErrors'; payload: Array<Error> };
```
dtfiedler, line 8:
```suggestion
```

src/state/state.tsx
```diff
@@ -1,8 +1,44 @@
+import React, { createContext, Dispatch, useContext, useReducer } from 'react';
 import type { ArNSContractState } from '../types';
+import type { Action } from './reducer';
+import testContractState from '../../data/contracts/bLAgYxAdX2Ry-nt6aH2ixgvJXbpsEYm28NgJgyqfs-U.json';
 
 export type GlobalState = {
   arnsSourceContract: ArNSContractState;
   gateway: string;
-  connectedWallet: string;
+  connectedWallet: any;
```
dtfiedler, line 9:
```suggestion
  connectedWallet?: string;
```

dtfiedler (approved):
looks good, will unlock a lot for us

PR #23, opened 2022-11-04
bug(fix): css fix on search bar

atticusofsparta:


src/components/inputs/Search/SearchBar/styles.css
```diff
@@ -4,13 +4,11 @@
...
   min-width: 250px;
   height: 65px;
-  padding-left: 26px;
-  border-radius: 2em;
   background-color: var(--text-white);
   border: 2px solid #e0e0e0;
   border-radius: 100px;
-  margin: auto;
   align-items: center;
+  padding-left: 26px;
```
dtfiedler, line 11: let's remove this, the input is the only thing that needs padding
```suggestion
```

src/index.css
```diff
@@ -54,8 +54,9 @@ body {
   font-style: normal;
   font-weight: 500;
   font-size: 12px;
-  width: 50%;
+  max-width: 400px;
```
dtfiedler, line 57:
```suggestion
  max-width: 400px;
  text-align: center;
```

src/index.css
```diff
@@ -54,8 +54,9 @@ body {
...
   margin: 0;
   justify-content: center;
   align-items: center;
 }
```
dtfiedler, line 61:
```suggestion
  max-width: 400px;
  text-align: center;
}
```

dtfiedler: we don't (and probably won't) have anything else in this div so we don't really need the justiy-content and align-items stuff. just the text alignment

dtfiedler (approved):

PR #29, opened 2022-11-09
feat(connect): add arns connect wallet modal

atticusofsparta:
image
image

### DIFF
--- a/src/components/icons/index.ts
+++ b/src/components/icons/index.ts
@@ -4,4 +4,4 @@ import ArweaveAppIcon from './ArweaveAppIcon.png';
 import UploadJsonKey from './UploadJsonKey.png';
 import { ReactComponent as CloseIcon } from './Close.svg';
 
-export { UploadJsonKey, SearchIcon, CloseIcon, ArweaveAppIcon, ArConnectIcon };
+export {ArConnectIcon, ArweaveAppIcon, CloseIcon, SearchIcon, UploadJsonKey};


src/types.ts
```diff
@@ -1,3 +1,5 @@
+import { Dispatch } from 'react';
```
dtfiedler, line 1: nit: needed?

atticusofsparta: nope, removed

src/components/icons/index.ts
```diff
@@ -1,3 +1,7 @@
 import { ReactComponent as SearchIcon } from './Search.svg';
+import ArConnectIcon from './ArConnectIcon.png';
+import ArweaveAppIcon from './ArweaveAppIcon.png';
+import UploadJsonKey from './UploadJsonKey.png';
+import { ReactComponent as CloseIcon } from './Close.svg';
 
-export { SearchIcon };
+export { SearchIcon, CloseIcon, ArConnectIcon, ArweaveAppIcon, UploadJsonKey };
```
dtfiedler, line 7: super nit: alphabatize

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <button className="walletConnectButton">
+          <img src={ArConnectIcon} alt="" width="47px" height="47px" />
+          Connect via ArConnect
+        </button>
+        <button className="walletConnectButton">
+          <img src={ArweaveAppIcon} alt="" width="47px" height="47px" />
+          Connect using Arweave.app
+        </button>
+        <text className="textFaded" style={{ marginTop: '1em' }}>
+          Don't have a wallet? get one{' '}
```
dtfiedler, line 53: it's a bit more clear to use `&nbsp;` then the bracketed space

src/components/modals/ConnectWalletModal/styles.css
```diff
@@ -0,0 +1,84 @@
...
+  box-shadow: 0px 0px 4px rgba(255, 255, 255, 0.25);
+  border-radius: 24px;
+  padding: 30px;
+  justify-content: center;
+  align-items: center;
+  row-gap: 1em;
+}
+.walletConnectButton {
+  display: flex;
+  width: 100%;
```
dtfiedler, line 62: nit: you don't need to set `width:100%` in flexbox with no other elements

dtfiedler (approved):
some non-obstructive nit picky things, otherwise looks good

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
atticusofsparta: this is actually from prettier - is their a config for this?

src/components/modals/ConnectWalletModal/styles.css
atticusofsparta: Apparently you do? I think the image/text inside counts as an element, just defaults to fit-content.

dtfiedler (approved):

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
dtfiedler: i think it'll add it if we don't explicitly add our own. also nit, these can be `spans` - not sure the difference with `text` tags

dtfiedler (approved):

PR #32, opened 2022-11-10
feat(regExp): Search bar name regex and error component

atticusofsparta:
image
Screenshot from 2022-11-10 11-01-22

### DIFF
--- a/src/components/inputs/Search/SearchBar/SearchBar.tsx
+++ b/src/components/inputs/Search/SearchBar/SearchBar.tsx
@@ -26,6 +26,9 @@ function SearchBar(props: SearchBarProps) {
   }
 
   function onHandleChange(e: React.ChangeEvent<HTMLInputElement>) {
+    if (e.target.value === '') {
+      reset();
+    }
     if (
       ARNS_NAME_REGEX.test(e.target.value) ||
       e.target.value.length < 1 ||
@@ -37,9 +40,7 @@ function SearchBar(props: SearchBarProps) {
     } else {
       setIsValid(false);
     }
-    if (e.target.value === '') {
-      reset();
-    }
+    
   }
 
   function onSubmit(name: string) {
@@ -104,19 +105,7 @@ function SearchBar(props: SearchBarProps) {
       </div>
       {React.cloneElement(footerElement, {
         ...props,
-        defaultText: (
-          <>
-            {!isValid ? (
-              <div className="errorContainer">
-                <span className="illegalChar">
-                  {footerElement.props.defaultText}
-                </span>
-              </div>
-            ) : (
-              footerElement.props.defaultText
-            )}
-          </>
-        ),
+        isValid,
         searchResult: submittedName
           ? { id: searchResult, domain: submittedName }
           : undefined,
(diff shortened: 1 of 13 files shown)


src/components/inputs/Search/SearchBar/SearchBar.tsx
```diff
@@ -17,30 +19,44 @@ function SearchBar(props: SearchBarProps) {
...
+        e.target.value.length > 1)
+    ) {
+      setSearchBarText(e.target.value);
+      setIsValid(true);
+    } else {
+      setIsValid(false);
+    }
     if (e.target.value === '') {
       reset();
     }
```
dtfiedler, line 42: put this first and return after, then you can remove the <1 parameters I believe

src/components/inputs/Search/SearchBar/SearchBar.tsx
```diff
@@ -49,15 +65,17 @@ function SearchBar(props: SearchBarProps) {
...
+        isAvailable,
       })}
       <div
         className="searchBar"
         style={
-          !searchSubmitted || isDefault
-            ? { borderColor: '' }
-            : isValid
-            ? { borderColor: 'var(--success-green)' }
+          isValid
```
dtfiedler, line 73:
```suggestion
          isAvailable
```

src/components/inputs/Search/SearchBar/SearchBar.tsx
```diff
@@ -84,6 +104,19 @@ function SearchBar(props: SearchBarProps) {
       </div>
       {React.cloneElement(footerElement, {
         ...props,
+        defaultText: (
+          <>
+            {!isValid ? (
```
dtfiedler, line 109: let's pass this to the `SearchBarFooter` element and put this div in there

src/components/inputs/Search/SearchBar/SearchBar.tsx
atticusofsparta: no still need that for if a user wants to completely delete the string. The second <1 is to allow dashes to be input on the end, which fails the regex, but is caught in the onSubmit function instead.

src/components/inputs/Search/SearchBar/SearchBar.tsx
atticusofsparta: added

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
```diff
@@ -49,14 +49,15 @@ function ConnectWalletModal({
           <img src={ArweaveAppIcon} alt="" width="47px" height="47px" />
           Connect using Arweave.app
         </button>
-        <text className="textFaded" style={{ marginTop: '1em' }}>
-          Don't have a wallet? get one{' '}
+        <text className="textWhite" style={{ marginTop: '1em' }}>
```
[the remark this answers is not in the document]
atticusofsparta: broke down some classes here, will continue to follow this pattern as I build more and come across them.

dtfiedler (approved):

PR #35, opened 2022-11-11
fix: some small changes and cleanup around search bar

dtfiedler:


atticusofsparta (approved):
made one change to a variable name to avoid conflict

PR #36, opened 2022-11-11
chore(eslint): setup eslint and rules

dtfiedler:
eslint is a static analysis tool that flags poor practices, types, etc. that don't follow the recommended patterns/rules described via various plugins. I've also added some plugins that sort imports and standardize file formats.


atticusofsparta (approved):
sounds like a good addition

PR #38, opened 2022-11-14
feat(tier cards): Arns upgrade tier

atticusofsparta:


types/constants.ts
```diff
@@ -1 +1,90 @@
...
+  24: 100000,
+  25: 100000,
+  26: 100000,
+  27: 100000,
+  28: 100000,
+  29: 100000,
+  30: 100000,
+  31: 100000,
+  32: 100000,
+};
```
dtfiedler, line 35: can't we get this from the global arns contract state?

atticusofsparta: yup. But that only has prices for up to 20 chars. Will need to add some logic for that - we could actually also implement a regEx updater that looks at how many prices their are in the contract and limits the regex to that. Thoughs?

src/types.ts
```diff
@@ -45,8 +46,14 @@ export type SearchBarFooterProps = {
...
   isSearchValid?: boolean;
+  isAvailable?: boolean;
 };
 
 export type ConnectWalletModalProps = {
   setShowModal: any;
 };
+export type TierCardProps = {
+  tier: number;
+  setTier: any;
```
dtfiedler, line 57:
```suggestion
  setTier: () => void;
```

types/constants.ts
```diff
@@ -1 +1,62 @@
...
+    'Up to 1,000 Subdomains',
+    'Available via all ArNS-enabled gateways',
+    'Permanently stored on Arweave',
+  ],
+  3: [
+    'Up to 10,000 Subdomains',
+    'Available via all ArNS-enabled gateways',
+    'Permanently stored on Arweave',
+  ],
+};
```
dtfiedler, line 19: i think future versions of the contract will have this on them, but fine for now

types/constants.ts
```diff
@@ -1 +1,62 @@
...
+export const NAME_PRICE_INFO =
+  'Registration fees are determined by the character length of the domain, lease duration, and what tier you choose.';
+
+export function NAME_PRICE_CALC({
+  domain,
+  years,
+  tier,
+  fees,
+}: {
+  domain: string | undefined;
```
dtfiedler, line 29: whenever you want to optionally define a param, you can use `?` instead of `| undefined`. 

```suggestion
  domain?: string;
```

types/constants.ts
```diff
@@ -1 +1,62 @@
...
+  years: number;
+  tier: number;
+  fees: { [x: number]: number };
+}) {
+  try {
+    if (years < 1) {
+      throw Error('must lease for atleast one year');
```
dtfiedler, line 36:
```suggestion
      throw Error('Minimum duration must be at least one year');
```

types/constants.ts
```diff
@@ -1 +1,62 @@
...
+    if (tier < 1) {
+      throw Error('Minimum tier is 1');
+    }
+    if (tier > 3) {
+      throw Error('Maximum tier is 3');
+    }
+    if (!domain) {
+      throw Error('Domain is undefined');
+    }
+    if (ARNS_NAME_REGEX.test(domain) && domain) {
```
dtfiedler, line 47: we should use our `validation` helper here instead of repeated logic

dtfiedler: also, follow the pattern above where we use the `false` scenario to throw errors, instead of if/else

src/components/inputs/YearsCounter/YearsCounter.tsx
```diff
@@ -0,0 +1,53 @@
...
+      `${date.toLocaleString('default', {
+        month: 'long',
+      })} ${date.getDate()}, ${newYear}`,
+    );
+  }
+  return (
+    <div className="yearsCounterContainer">
+      <p className="text white bold">Registration Period</p>
+      <div className="yearsCounter">
+        <button className="counterButton" onClick={() => subCount()}>
```
dtfiedler, line 41: it's good to be explicit in function names `subtractFromYears` or something similar

src/components/inputs/YearsCounter/YearsCounter.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  function subCount() {
+    if (count > 1) {
+      setCount(count - 1);
+    } else {
+      return;
+    }
+  }
+  function addCount() {
+    setCount(count + 1);
```
dtfiedler, line 21: the contract will probably define this, but we should have a max-count for years. at least for now. maybe 5/10? what do you think?

src/components/layout/UpgradeTier/UpgradeTier.tsx
```diff
@@ -0,0 +1,64 @@
...
+    setPrice(NAME_PRICE_CALC({ domain, tier, years, fees }));
+  }, [years, tier, domain, arnsSourceContract]);
+
+  return (
+    <div className="upgradeTier">
+      <YearsCounter setCount={setYears} count={years} />
+      <div className="cardContainer">
+        <TierCard thisTier={1} setTier={setTier} tier={tier} />
+        <TierCard thisTier={2} setTier={setTier} tier={tier} />
+        <TierCard thisTier={3} setTier={setTier} tier={tier} />
```
dtfiedler, line 29: i think `selectedTier` and `tier` are a little more clear
```suggestion 
      // TODO: pul tiers from contract when evolved
       Object.keys(TIER_DATA).map((tier) => {
            return  <TierCard tier={tier} setTier={setTier} selectdTier={selectedTier} />
```

src/components/layout/UpgradeTier/UpgradeTier.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useStateValue } from '../../../state/state';
+import TierCard from '../../cards/TierCard/TierCard';
+import { AlertCircle } from '../../icons';
+import YearsCounter from '../../inputs/YearsCounter/YearsCounter';
+import './styles.css';
+
+function UpgradeTier({ domain }: { domain: string | undefined }) {
+  const [{ arnsSourceContract }] = useStateValue();
+  // name is passed down from search bar to calculate price
+  const [tier, setTier] = useState(1);
```
dtfiedler, line 13:
```suggestion
  const [selectedTier, setSelectedTier] = useState(1);
```

src/types.ts
```diff
@@ -45,8 +46,14 @@ export type SearchBarFooterProps = {
...
+  thisTier: number;
```
dtfiedler, line 58:
```suggestion
  selectedTier: number;
```

types/constants.ts
```diff
@@ -1 +1,62 @@
...
+  3: [
+    'Up to 10,000 Subdomains',
+    'Available via all ArNS-enabled gateways',
+    'Permanently stored on Arweave',
+  ],
+};
+export const NAME_PRICE_INFO =
+  'Registration fees are determined by the character length of the domain, lease duration, and what tier you choose.';
+
+export function NAME_PRICE_CALC({
```
dtfiedler, line 23: nit: put in the `utils` directory, and name it `calculateArNSDomainPrice`

types/constants.ts
```diff
@@ -1 +1,62 @@
...
+      const namePrice = fees[nameLength];
+      const price = namePrice * years * tier;
+
+      return price;
+    } else {
+      throw Error('invalid name, failed name regex');
+    }
+  } catch (error) {
+    return 0;
+  }
```
dtfiedler, line 61:
```suggestion
  } catch (error) {
    // TODO: when invalidated price, should we throw error?
    console.error(error)
    return 0;
  }
```

dtfiedler: i'm thinking about how we want to disable the purchasing if an error is thrown, perhaps we don't catch the error and make the calling parent have to handle it

src/components/inputs/YearsCounter/YearsCounter.tsx
atticusofsparta: Actually I had the same question, currently the opinion from the others was no max count. Ill ping in slack for another take from phil/johnathan though.

types/constants.ts
atticusofsparta: yeah good point. Following logic for global state from price-calculator repo, they have a global state for errors.... maybe even a global listener for them, not sure. Anyways we can push the error to our global error object for that i think.

types/constants.ts
dtfiedler: 20-32 is the same price, so if the length is greater than 20, use that price

src/components/inputs/YearsCounter/YearsCounter.tsx
dtfiedler: also, FWIW - the outputs of this convo will be out of scope for this ticket, we'll create a new one to adjust based on the feedback from the group

src/components/layout/UpgradeTier/UpgradeTier.tsx
```diff
@@ -0,0 +1,64 @@
+import { useEffect, useState } from 'react';
+
+import { NAME_PRICE_CALC, NAME_PRICE_INFO } from '../../../../types/constants';
+import { useStateValue } from '../../../state/state';
+import TierCard from '../../cards/TierCard/TierCard';
+import { AlertCircle } from '../../icons';
+import YearsCounter from '../../inputs/YearsCounter/YearsCounter';
+import './styles.css';
+
+function UpgradeTier({ domain }: { domain: string | undefined }) {
```
dtfiedler, line 10:
```suggestion
function UpgradeTier({ domain }: { domain?: string }) {
```

src/components/inputs/YearsCounter/YearsCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+  return (
+    <div className="yearsCounterContainer">
+      <p className="text white bold">Registration Period</p>
+      <div className="yearsCounter">
+        <button className="counterButton" onClick={() => subtractYearsCount()}>
+          -
+        </button>
+        <p className="text bold">{`${count} ${years}`}</p>
+        <button className="counterButton" onClick={() => addYearsCount()}>
```
dtfiedler, line 48: nit: disable when count < MAX_COUNT

src/components/inputs/YearsCounter/YearsCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+      `${date.toLocaleString('default', {
+        month: 'long',
+      })} ${date.getDate()}, ${newYear}`,
+    );
+  }
+  return (
+    <div className="yearsCounterContainer">
+      <p className="text white bold">Registration Period</p>
+      <div className="yearsCounter">
+        <button className="counterButton" onClick={() => subtractYearsCount()}>
```
dtfiedler, line 44: disable when count == MIN_YEARS_COUNT (e.g. 1)

src/components/inputs/YearsCounter/YearsCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+    }
+    const newYear = date.getFullYear() + count;
+    setRegistration(
+      `${date.toLocaleString('default', {
+        month: 'long',
+      })} ${date.getDate()}, ${newYear}`,
+    );
+  }
+  return (
+    <div className="yearsCounterContainer">
```
dtfiedler, line 41: just FYI - we have a lot of component specific CSS happening. we should continue to try and create global classes and re-use them. not going to block this PR on this but something to think about when adding new components

src/components/inputs/YearsCounter/YearsCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+    setCount(count + 1);
+  }
+  function changeYear() {
+    const date = new Date();
+    if (count > 1) {
+      setYears(`years`);
+    } else {
+      setYears(`year`);
+    }
+    const newYear = date.getFullYear() + count;
```
dtfiedler, line 33: nit: you can do something like `Date.setYear`  to manipulate the actual date

src/components/inputs/YearsCounter/YearsCounter.tsx
dtfiedler: cool, probably worht setting as a constant with //todo to address it when that changes

src/components/inputs/YearsCounter/YearsCounter.tsx
```diff
@@ -0,0 +1,56 @@
+import { useEffect, useState } from 'react';
+
+import './styles.css';
+
+function YearsCounter({ setCount, count }: { setCount: any; count: number }) {
```
dtfiedler, line 5:
```suggestion
function YearsCounter({ setCount, count }: { setCount: () => void; count: number }) {
```

src/components/cards/TierCard/TierCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { TIER_DATA } from '../../../../types/constants';
+import { TierCardProps } from '../../../types';
+import { CircleCheck } from '../../icons';
+import './styles.css';
+
+function TierCard({ tier, setTier, selectedTier }: TierCardProps) {
+  const [thisTier, setThisTier] = useState(new Number(tier).valueOf());
+
+  useEffect(() => {
+    setThisTier(new Number(tier).valueOf());
```
dtfiedler, line 12: why do we need to do this casting here if a `number` is passed?

src/utils/searchUtils.ts
```diff
@@ -21,3 +21,47 @@ export function isArNSDomainNameAvailable({
...
+    if (isArNSDomainNameValid({ name: domain }) && domain) {
+      let nameLength = domain.length;
+      if (nameLength > Object.keys(fees).length) {
+        nameLength = Object.keys(fees).length;
+      }
+      const namePrice = fees[nameLength];
+      const price = namePrice * years * thisTier;
+
+      return price;
+    }
```
dtfiedler, line 62: ah, wasn't clear. i meant throw an error, but avoid the use of `else.
```suggestion
    if (!isArNSDomainNameValid({ name: domain })) {
      throw Error('Domain name is invalid');
    }
      const nameLength = Math.max(domain.length, Object.keys(fees).length);
      const namePrice = fees[nameLength];
      const price = namePrice * years * thisTier;

      return price;
    }
```

src/components/cards/TierCard/TierCard.tsx
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from 'react';
+
+import { TIER_DATA } from '../../../../types/constants';
+import { TierCardProps } from '../../../types';
+import { CircleCheck } from '../../icons';
+import './styles.css';
+
+function TierCard({ tier, setTier, selectedTier }: TierCardProps) {
+  const [thisTier, setThisTier] = useState(new Number(tier).valueOf());
```
dtfiedler, line 9: why can't we just use `tier`? it's set by the parent, and no changes internally

atticusofsparta: cant access it while tier is a string: because its accessing it via object.keys its returning it as a string. This pattern takes it wether is a string or a number an creates a number from it. I could probably move the pattern to the .map function in upgrade tier instead... thoughts?

src/components/inputs/YearsCounter/YearsCounter.tsx
atticusofsparta: this isnt actually working for the useState functions... it recommends something like Dispatch<SetState<inffered type from default state>>

src/components/inputs/YearsCounter/YearsCounter.tsx
atticusofsparta: note on min/max lease durations: can be set in seconds or milliseconds for dynamic changes (instead of years for example). ex if we say 1 year is the minumum, we put it in ms (31_536_000_000) instead of "1" so that if we change the minimum lease period's count we can leave the logic as-is.

src/types.ts
```diff
@@ -45,8 +48,14 @@ export type SearchBarFooterProps = {
...
   searchResult?: ArNSDomain;
   isSearchValid?: boolean;
+  isAvailable?: boolean;
 };
 
 export type ConnectWalletModalProps = {
   setShowModal: any;
 };
+export type TierCardProps = {
+  tier?: number | string;
```
dtfiedler, line 58: why is this optionally a string?

src/components/layout/UpgradeTier/UpgradeTier.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(() => {
+    const fees = arnsSourceContract.fees;
+    setPrice(calculateArNSNamePrice({ domain, selectedTier, years, fees }));
+  }, [years, selectedTier, domain, arnsSourceContract]);
+
+  return (
+    <div className="upgradeTier">
+      <YearsCounter setCount={setYears} count={years} />
+      <div className="cardContainer">
+        {Object.keys(TIER_DATA).map((tier, index) => (
```
dtfiedler, line 28:
```suggestion
        {Object.keys(TIER_DATA).map((tier: number, index: number) => (
```

src/components/inputs/YearsCounter/YearsCounter.tsx
dtfiedler: let's just allow years for now

src/components/cards/TierCard/TierCard.tsx
dtfiedler: if we limit the `tier` to only be a number on the type, then make sure we cast it before hand - does that solve this problem? regardless, we should force the parent to cast it to a number and not allow anything else

dtfiedler (approved):

dtfiedler (approved):
lg - nice work!

PR #40, opened 2022-11-15
fix: some home page resizing, adjustments to nav bar component, etc.

dtfiedler:


atticusofsparta (dismissed):
some tests were failing because an auto-import issue (added js extension) removed the auto import and fixed it.

PR #47, opened 2022-11-22
fix: follow up from demo, adding registration buttons, mobile nav bars, etc.

dtfiedler:
Mobile:
image
image


src/utils/constants.ts
```diff
@@ -1,5 +1,7 @@
 // export const ARNS_NAME_REGEX = new RegExp('^(?!-)[a-z0-9-s+]{1,32}(?<!-)$');
-export const ARNS_NAME_REGEX = new RegExp('^[a-z0-9-s+]{1,32}$');
+export const ARNS_NAME_REGEX = new RegExp(
+  '^[a-zA-Z0-9][a-zA-Z0-9-]{0,30}[a-zA-Z0-9]$',
```
atticusofsparta, line 3: change to '^([a-zA-Z0-9][a-zA-Z0-9-]{0,30}[a-zA-Z0-9]|[a-zA-Z0-9]{1})$' to allow for single char names

src/components/inputs/Counter/Counter.tsx
```diff
@@ -0,0 +1,108 @@
...
+          >
+            -
+          </button>
+          <input
+            className="counterInput text bold"
+            type="number"
+            value={count}
+            pattern={'/^[1-9]{1,3}$/'}
+            onFocus={(e) => e.target.select()}
+            onChange={(e) =>
```
atticusofsparta, line 87: this sets the MIN_LEASE_COUNT but not the max (allows more than the max to be entered) recommend changing to `onChange={(e) =>
              +e.target.value > MIN_LEASE_DURATION ? setCount(Math.min(MAX_LEASE_DURATION, +e.target.value))
              : setCount(MIN_LEASE_DURATION)
            }`

atticusofsparta (approved):

PR #58, opened 2022-12-18
feat(wallets): fetch AR balance from arweave

dtfiedler:
### Details

Adds `getWalletBalanceAR` to the `ArweaveWalletConnector` interface and implements it on the `ArConnectWalletConnector`. Also adds an `arweave` client to the global state, that updates when gateway changes.

#### Desktop
image

##### Mobile
image


atticusofsparta (commented):
nit: format of balances / address should match figma 
image

atticusofsparta (commented):
nit: noticing some rendering stutters when opening the menu, not sure what is causing it but would probably be a good idea to come back and change how its rendering

atticusofsparta (approved):
looks good

PR #75, opened 2023-01-10
feat: update design, nav bar and routes

dtfiedler:


src/components/layout/Navbar/NavBarLink/styles.css
```diff
@@ -6,4 +6,5 @@
   font-size: 16px;
   font-family: 'Rubik-Bold';
   text-decoration: none;
+  opacity: 50%;
```
atticusofsparta, line 9: Should be 100% by default according to figma unless a menu is open

src/components/inputs/buttons/MenuButton/MenuButton.tsx
```diff
@@ -5,19 +5,26 @@ import './styles.css';
...
-  icon: JSX.Element;
+  children?: JSX.Element;
 }): JSX.Element {
   return (
     <>
       <button
-        className={!show ? 'round-button' : 'round-button white-background'}
+        className={
+          !show
+            ? 'outline-button hover'
```
atticusofsparta, line 19: the profile button differs from the outline button: figma shows 2px border width and 6 px radius

atticusofsparta:
Since we are updating the nav menu card we could add the icon for Log Out here. You can find the correct icon here: https://www.figma.com/file/na4spGwQ0RfniQdbh8LKJT/ar.io-Design-System?node-id=474%3A3&t=fwYbgnM05MRGX436-0
image

src/components/inputs/buttons/MenuButton/MenuButton.tsx
dtfiedler: updated

src/components/layout/Navbar/NavBarLink/styles.css
dtfiedler: turned it back to 50%, will add some logic later for when the menu is open

atticusofsparta (approved):

PR #76, opened 2023-01-10
Feat: Fetch deployed ants

atticusofsparta:
fetches ants deployed by connected user and hands them to manage-assets page.
future work will implement this in manage-assets branch to paginate using the provided cursor


src/services/wallets/ArConnectWalletConnector.ts
```diff
@@ -44,4 +53,36 @@ export class ArConnectWalletConnector implements ArweaveWalletConnector {
...
+          if (ant.cursor) {
+            newCursor = ant.cursor;
+          }
+        });
+    }
+    return {
+      ids: [...fetchedANTids],
+      cursor: newCursor,
+    };
+  }
```
dtfiedler, line 87: since these queries will likely be the same across all connectors we can create an abstract class called `GraphQLAPI` and separate the interfaces.

```
export interface ArweaveWalletConnector {
  connect(): Promise<void>;
  disconnect(): Promise<void>;
  getWalletAddress(): Promise<string>;
  getWalletBalanceAR(): Promise<string>;
}

export interface ArweaveGraphQLAPI {
  getWalletANTs(
    approvedSourceCodeTransactions: string[],
    cursor?: string,
  ): Promise<{ ids: string[]; cursor?: string }>;
}
```
and then implement the queries in an abstract class 

```
abstract class ArweaveGraphQL implements ArweaveGraphQLAPI {
  constructor(arweave: Arweave) {
      this._arweave = arweave
  }
    async getWalletANTs(
    approvedSourceCodeTransactions: string[],
    cursor?: string,
  ): Promise<{ ids: string[]; cursor?: string }> {
    const address = await this.getWalletAddress();
    const fetchedANTids: Set<string> = new Set();
    let newCursor: string | undefined = undefined;

    // get contracts deployed by user, filtering with src-codes to only get ANT contracts
    const deployedResponse = await this._arweave.api.post(
      '/graphql',
      deployedAntQuery(address, approvedSourceCodeTransactions, cursor),
    );
    if (deployedResponse.data.data?.transactions?.edges?.length) {
      deployedResponse.data.data.transactions.edges
        .map((e: any) => ({
          id: e.node.id,
          cursor: e.cursor,
        }))
        .forEach((ant: { id: string; cursor: string }) => {
          fetchedANTids.add(ant.id);
          if (ant.cursor) {
            newCursor = ant.cursor;
          }
        });
    }
    return {
      ids: [...fetchedANTids],
      cursor: newCursor,
    };
  }
```

and then have the `ArConnectWalletConnector` extend  `ArweaveGraphQL`

export class ArConnectWalletConnector implements ArweaveWalletConnector extends ArweaveGraphQL {
  ...attributes
  constructor(arweave: Arweave) {
     super()
  }
 ...
}
```
also note: we pass `arweave` into the constructor, rather than relying on the static client since we will likey give users to options to change gateways.

File: src/services/wallets/JsonWalletConnector.ts [hunk 2]
```diff
@@ -1,15 +1,23 @@
...
+import { deployedAntQuery } from '../../utils/constants';
+import { arweave } from '../arweave/arweave';
 
 // A lot to do here, r.e. security, we will likely move to a different approach.
 export class JsonWalletConnector implements ArweaveWalletConnector {
   private _walletFile;
   private _wallet?: JWKInterface;
+  private _arweave: Arweave;
+  private _ar: Ar = new Ar();
+  private _address?: string;
```
dtfiedler, line 16: are these needed? they don't seem to be used anywhere

src/services/wallets/ArConnectWalletConnector.ts
dtfiedler: this approach is `composition` - another approach is `inheritance` where we create a super class, that `ArConnectWalletConnector` can extend, but that's not generally favored when trying to achieve code re-use. a good write up here: https://www.geeksforgeeks.org/favoring-composition-over-inheritance-in-java-with-examples/

src/services/wallets/JsonWalletConnector.ts
atticusofsparta: _ar is used for parsing balance to winston and address is used in the getAnts query - they are needed.

dtfiedler: they aren't _yet_ used in this class though 😉

atticusofsparta: ohhhh i didnt read the file name, youre right they can be removed. i think.

src/types.ts
```diff
@@ -81,6 +81,18 @@ export interface ArweaveWalletConnector {
   disconnect(): Promise<void>;
   getWalletAddress(): Promise<string>;
   getWalletBalanceAR(): Promise<string>;
+  getWalletANTs(
+    approvedSourceCodeTransactions: string[],
+    cursor?: string,
+  ): Promise<{ ids: string[]; cursor?: string }>;
```
dtfiedler, line 87:
```suggestion
```

dtfiedler: we also should update the type of the global state so the wallet is:
```
export type GlobalState = {
  arweave: Arweave;
  arnsSourceContract: ArNSContractState;
  gateway: string;
  walletAddress?: string;
  wallet?: ArweaveWalletConnector & ArweaveGraphQLAPI;
  arnsContractId: string;
  showConnectWallet: boolean;
  errors: Array<Error>;
  notifications: { id: string; text: string }[];
};
```
and the reducer
```
export type Action =
  | { type: 'setWalletAddress'; payload: ArweaveTransactionId | undefined }
  | {
      type: 'setWallet';
      payload: (ArweaveWalletConnector & ArweaveGraphQLAPI) | undefined;
    }
  | { type: 'setGateway'; payload: string }
  | { type: 'setArnsContractState'; payload: ArNSContractState }
  | { type: 'setShowConnectWallet'; payload: boolean }
  | { type: 'setArweave'; payload: Arweave }
  | { type: 'pushNotification'; payload: string }
  | { type: 'removeNotification'; payload: string };
```

src/services/wallets/JsonWalletConnector.ts
```diff
@@ -58,4 +66,35 @@ export class JsonWalletConnector implements ArweaveWalletConnector {
...
+          if (ant.cursor) {
+            newCursor = ant.cursor;
+          }
+        });
+    }
+    return {
+      ids: [...fetchedANTids],
+      cursor: newCursor,
+    };
+  }
```
dtfiedler, line 99:
```suggestion
```

atticusofsparta: will remove the majority but will have to stub this after updating the global wallet type i think. Unless we want to remove the json connector completely.

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
```diff
@@ -98,7 +99,10 @@ function ConnectWalletModal({ show }: { show: boolean }): JSX.Element {
         </button>
         <button
           className="wallet-connect-button h2"
-          onClick={() => setGlobalWallet(new ArConnectWalletConnector(arweave))}
+          onClick={() => {
+            const graphql = new ArweaveGraphQL(arweave);
+            setGlobalWallet(new ArConnectWalletConnector(arweave, graphql));
```
dtfiedler, line 104:
```suggestion
            setGlobalWallet(new ArConnectWalletConnector(arweave));
```

src/services/wallets/JsonWalletConnector.ts
dtfiedler: you shouldn't have to stub it since `getWalletAnts` is no longer in the `ArweaveWalletConnector` interface

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
atticusofsparta: fixed

src/services/wallets/JsonWalletConnector.ts
atticusofsparta: removed JsonWalletConnector

dtfiedler: we don't need to delete the class, just remove it from the connect modal and anywhere else it's used.

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
```diff
@@ -77,28 +71,12 @@ function ConnectWalletModal({ show }: { show: boolean }): JSX.Element {
...
-            <input
-              className="hidden"
-              type="file"
-              onChange={(e) =>
-                e.target?.files?.length &&
-                setGlobalWallet(new JsonWalletConnector(e.target.files[0]))
-              }
-            />
-          </label>
-        </button>
```
dtfiedler, line 98: remove just this, lets keep the class - we may want to use it later for testing

dtfiedler (changes requested):
add back `JsonWalletConnector` class (keep it removed from the modal) and then it should be good

dtfiedler (approved):
lg!

PR #78, opened 2023-01-11
chore: Warp update to L1 settings

atticusofsparta:


src/services/arweave/WarpDataProvider.ts
```diff
@@ -9,8 +10,14 @@ import {
...
-  constructor() {
-    this._warp = WarpFactory.forMainnet();
+  constructor(arweave: Arweave) {
+    this._warp = WarpFactory.forMainnet(
+      {
+        ...defaultCacheOptions,
+      },
+      true,
+      arweave,
+    );
```
dtfiedler, line 20: nit: please add a comment here as to why we are using the arweave gateway

atticusofsparta: added comment

dtfiedler (dismissed):

dtfiedler (approved):

PR #90, opened 2023-01-21
chore: Transaction id type

atticusofsparta:


types/ArweaveTransactionID.ts
```diff
@@ -0,0 +1,44 @@
...
+    return this.transactionId;
+  }
+
+  equals(entityId: ArweaveTransactionID): boolean {
+    return this.transactionId === entityId.transactionId;
+  }
+
+  toJSON(): string {
+    return this.toString();
+  }
```
dtfiedler, line 35: do we need these?

types/ArweaveTransactionID.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function TxID(transactionId: string): ArweaveTransactionID {
+  return new ArweaveTransactionID(transactionId);
+}
```
dtfiedler, line 40: nit: if probably not necessary - we can just use the constructor

types/ArweaveTransactionID.ts
atticusofsparta: Probably not, can take them out if you want, but I see no reason to not leave them in. I tried looking thru the ardrive core to see where they are used but its proving difficults. I can see myself using `.equals()` to check if a transaction is in the registry or if they are the owner of a contracts. toJSON()....im not sure it useful - theres this line `	[Symbol.toPrimitive](hint?: string): string {
		if (hint === 'number') {
			throw new Error('Transaction IDs cannot be interpreted as a number!');
		}

		return this.toString();
	}` that seems to automatically turn the txid into a string... what do you think?

types/ArweaveTransactionID.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const stubTransactionID = TxID(
+  '0000000000000000000000000000000000000000000',
+);
```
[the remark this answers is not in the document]
atticusofsparta: moved

types/ArweaveTransactionID.ts
atticusofsparta: removed

types/ArweaveTransactionID.ts
dtfiedler: i think leave `equals` and `toString` - we can always add more later but I don't anticipate needing them

types/ArweaveTransactionID.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+
+  [Symbol.toPrimitive](hint?: string): string {
+    if (hint === 'number') {
+      throw new Error('Transaction IDs cannot be interpreted as a number!');
+    }
+
+    return this.toString();
+  }
```
dtfiedler, line 19: i don't believe we'll ever run into this issue since we don't try and case transaction ids to numbers. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive)

the error would show if you did something like
````
const id = new ArweaveTransactionID('gh673M0Koh941OIITVXl9hKabRaYWABQUedZxW-swIA')
const idAsNumber = +id; // ERROR WOULD THROW ON THIS CASTING
```

atticusofsparta: actually, i think i have run into this issue, but i cant remember when. Transaction ids that started with a number threw a type error. Believe it was something to do with the gql query and the user wallet.

types/ArweaveTransactionID.ts
atticusofsparta: removed

dtfiedler (approved):

PR #105, opened 2023-02-10
fix: introduce transaction types for headers and tags

dtfiedler:


src/services/arweave/SimpleArweaveDataProvider.ts
```diff
@@ -25,20 +30,18 @@ export class SimpleArweaveDataProvider implements ArweaveDataProvider {
...
-      .then((res: any) => {
-        return res.data.number_of_confirmations;
-      });
-    return confirmations;
+    const { status, data } = await this._arweave.api.get(`/tx/${id}/status`);
+    if (status !== 200) {
+      throw Error('Failed fetch confirmations for transaction id.');
+    }
+    return +data.number_of_confirmations;
   }
```
dtfiedler, line 38: the arweave client swallows all errors (404's, 429's etc.) so we want to to check the `status` and throw our own error

src/services/arweave/SimpleArweaveDataProvider.ts
```diff
@@ -133,6 +141,7 @@ export class SimpleArweaveDataProvider implements ArweaveDataProvider {
     }
   }
   async validateArweaveId(id: string): Promise<ArweaveTransactionID> {
+    // todo: we should change this to a promise that checks for the existing of the tx id instead of just constructing a promise that rejects on class creation
```
dtfiedler, line 144: this could be as easy as callling `this.getTransactionHeaders(id)` and validating the status returns a `200`

atticusofsparta (approved):

PR #112, opened 2023-02-16
Register name fix

atticusofsparta:


dtfiedler (approved):
let's test this as soon as it goes out

PR #118, opened 2023-02-24
PE-3097: extend interface for deployContract function

atticusofsparta:


src/services/arweave/WarpDataProvider.ts
```diff
@@ -79,4 +96,97 @@ export class WarpDataProvider implements SmartweaveDataProvider {
...
+    stateType,
+    tags = [],
+  }: {
+    srcCodeTransactionId: ArweaveTransactionID;
+    initialState: ANTContractState | ArweaveTransactionID;
+    stateType: 'TXID' | 'TAG' | 'DATA';
+    tags: TransactionTag[];
+  }): Promise<string | undefined> {
+    const stateSize = new Blob([JSON.stringify(initialState)]).size;
+    const tagSize = new Blob([JSON.stringify(tags)]).size;
```
dtfiedler, line 118: might be helpful as a `util` function

src/services/arweave/WarpDataProvider.ts
```diff
@@ -79,4 +96,97 @@ export class WarpDataProvider implements SmartweaveDataProvider {
...
+        );
+      }
+      if (
+        stateSize + tagSize > SMARTWEAVE_MAX_TAG_SPACE &&
+        stateType !== 'TXID'
+      ) {
+        throw new Error(
+          `Combined size of tags (${tagSize} bytes) and state (${stateSize} bytes) too large for tag space, must be under ${SMARTWEAVE_MAX_TAG_SPACE} bytes. Reduce the tags or use a different state deployment method.`,
+        );
+      }
```
dtfiedler, line 149: do you mind adding a comment with the Smartweave docs that speak to these size limits?

src/services/arweave/WarpDataProvider.ts
```diff
@@ -79,4 +96,97 @@ export class WarpDataProvider implements SmartweaveDataProvider {
...
+        wallet: ArWallet;
+        initState: string;
+        srcTxId: string;
+        tags: TransactionTag[];
+      } = {
+        wallet: 'use_wallet',
+        initState: '',
+        srcTxId: srcCodeTransactionId.toString(),
+        tags: tags,
+      };
```
dtfiedler, line 160: nit: prefer to use `Map` when we are creating an object and dynamically updating it's values in these classes

src/services/arweave/LocalFilesystemDataProvider.ts
```diff
@@ -37,4 +42,22 @@ export class LocalFileSystemDataProvider implements SmartweaveDataProvider {
...
+    initialState: ArweaveTransactionID | ANTContractState;
+    stateType: 'TXID' | 'TAG' | 'DATA';
+    tags: TransactionTag[];
+  }): Promise<string | undefined> {
+    return JSON.stringify([
+      srcCodeTransactionId,
+      initialState,
+      stateType,
+      tags,
+    ]);
```
dtfiedler, line 61: why is it an array?

src/services/arweave/WarpDataProvider.ts
dtfiedler: also, why are we casting it to an Array? we could also use buffers, but would likely need a polyfill for the browser

src/utils/constants.ts
```diff
@@ -94,3 +94,5 @@ export const approvedContractsForWalletQuery = (
   };
   return queryObject;
 };
+
+export const SMARTWEAVE_MAX_TAG_SPACE = 2048 - 250; // minimum tag size of smartweave tags from warp is 239, rounded it for wiggle room
```
dtfiedler, line 98: nit: add link to smartweave docs for this

src/services/arweave/WarpDataProvider.ts
atticusofsparta: we are already polyfilling buffer due to other deps so we can if its better

dtfiedler: `Buffer.byteLength(JSON.stringify(initialState)`; or `Buffer.from(JSON.stringify(initalState).length`

dtfiedler:
we should consider adding unit tests to these interfaces and using arlocal - I'll create a ticket to set up the initial pattern that we can continue to extend. we DO NOT want to introduce this then create potential regressions due to poor testing

PR #120, opened 2023-02-25
chore(PE-3102) - ant contract class

atticusofsparta:


src/services/arweave/AntContract.tsx
```diff
@@ -0,0 +1,299 @@
...
+  ANTContractState,
+  ArweaveDataProvider,
+  ArweaveTransactionID,
+  SmartweaveDataProvider,
+} from '../../types';
+
+export class AntContract {
+  private _provider: SmartweaveDataProvider & ArweaveDataProvider;
+  state?: ANTContractState = undefined;
+  antId;
```
dtfiedler, line 11:
```suggestion
  antId: ArweaveTransactionID?;
```

src/services/arweave/AntContract.tsx
```diff
@@ -0,0 +1,299 @@
...
+   */
+  constructor(
+    walletAddress: ArweaveTransactionID,
+    provider: SmartweaveDataProvider & ArweaveDataProvider,
+    antId?: ArweaveTransactionID,
+  ) {
+    this.walletAddress = walletAddress;
+    this.antId = antId;
+    this._provider = provider;
+    this.newAntState = {
```
dtfiedler, line 29: nit: can we use just `state` and updates happen there. i don't see the need for two.

src/services/arweave/AntContract.tsx
```diff
@@ -0,0 +1,299 @@
+import {
+  ANTContractState,
+  ArweaveDataProvider,
+  ArweaveTransactionID,
+  SmartweaveDataProvider,
+} from '../../types';
+
+export class AntContract {
+  private _provider: SmartweaveDataProvider & ArweaveDataProvider;
+  state?: ANTContractState = undefined;
```
dtfiedler, line 10: this might be a good chance to use a `Map` given we will be writing/rewriting fields. i know warp doesn't play nice with raw maps, so we could introduce a `toJSON` function as well

src/services/arweave/AntContract.tsx
atticusofsparta: newAntState is a constant that is built using the walletAddress the class is given - changes dont happen there. Maybe DEFAULT_STATE would be a better name.

dtfiedler: can't we just set `this.state` to the default state before it loads? do we need two?

dtfiedler:
since these is _independent_ of `deployContract` logic, can we make it directly against develop branch? ideally PR's don't become deps of each other (it happens)

src/services/arweave/AntContract.tsx
```diff
@@ -0,0 +1,299 @@
...
+    }
+  }
+
+  async update() {
+    return;
+    // for batch interaction function, could generate a diff of state keys, then write an interaction for each addition.
+    // so check if new owner !== old owner, if not equal, write a newOwner interaction. If an undername was added write that... etc
+    // this would enable the user to batch edit a bunch of keys in the contract, then save the contract state, and the app
+    // would write all the interactions automatically. UI could have an Undo and Reset function to return to previous states
+  }
```
dtfiedler, line 298: nit: remove or comment out for now

src/services/arweave/AntContract.tsx
dtfiedler: also, we could elevate the attributes of state to getters/setters on this class - what do you think?
```
class foo {
    private _bar: boolean = false;
    get bar(): boolean {
        return this._bar;
    }
    set bar(value: boolean) {
        this._bar = value;
    }
}
```

src/services/arweave/AntContract.tsx
```diff
@@ -0,0 +1,262 @@
...
+    return this.contract.owner;
+  }
+  set owner(id: string) {
+    this.contract.owner = id;
+  }
+  get name() {
+    return this.contract.name;
+  }
+  set name(name: string) {
+    this.contract.name = name;
```
dtfiedler, line 229: any validations we want to do on these fields?

src/services/arweave/AntContract.tsx
```diff
@@ -0,0 +1,262 @@
...
+  records: {
+    '@': {
+      transactionId: '',
+      ttlSeconds: 1800,
+      maxSubdomains: 100,
+    },
+  },
+};
+
+export class CompositeAntContractState {
```
dtfiedler, line 24: FWIW we don't need this for `create an ANT` workflow

src/services/arweave/AntContract.tsx
```diff
@@ -0,0 +1,262 @@
...
+  set controller(controller: string) {
+    // any validations we want to do on the controller
+    this.contract.controller = controller;
+  }
+  get evolve() {
+    return this.contract.evolve;
+  }
+  set evolve(evolve: boolean | undefined) {
+    this.contract.evolve = evolve;
+  }
```
dtfiedler, line 255: let's exclude evolve for now

dtfiedler (approved):

PR #121, opened 2023-03-01
feat(interface): added deployContract function to interface

dtfiedler:


src/types.ts
```diff
@@ -90,6 +90,17 @@ export interface SmartweaveDataProvider {
...
   ): Promise<number>;
+  deployContract({
+    srcCodeTransactionId,
+    initialState,
+    stateType, // 'TXID', 'TAG', 'DATA' - configures how to apply the state to the transaction
+    tags,
+  }: {
+    srcCodeTransactionId: ArweaveTransactionID;
+    initialState: ANTContractJSON | ArweaveTransactionID;
+    stateType: 'TXID' | 'TAG' | 'DATA';
```
dtfiedler, line 101: within the ArNS Portal - we'll likely only be using `DATA`  -  do we need to support these other methods? i know they are available in smartweave but out of scope for how we'll deploy contracts

src/services/arweave/WarpDataProvider.ts
```diff
@@ -79,4 +96,97 @@ export class WarpDataProvider implements SmartweaveDataProvider {
...
+          ...deploymentPayload.tags,
+        ];
+      }
+      if (stateType === 'DATA') {
+        deploymentPayload.initState = JSON.stringify(initialState);
+      }
+      const { contractTxId } = await this._warp.deployFromSourceTx(
+        deploymentPayload,
+        true,
+      );
```
dtfiedler, line 180: see question below - is it necessary we support `TXID` and `TAG`?

src/services/arweave/WarpDataProvider.ts
```diff
@@ -50,9 +61,15 @@ export class WarpDataProvider implements SmartweaveDataProvider {
     },
   ): Promise<ArweaveTransactionID | undefined> {
     try {
+      const payloadSize = new Blob([JSON.stringify(payload)]).size;
```
dtfiedler, line 64: nit: buffers - `Buffer.byteLength(JSON.stringify(payload));`

src/services/arweave/WarpDataProvider.ts
```diff
@@ -79,4 +96,97 @@ export class WarpDataProvider implements SmartweaveDataProvider {
     const state = await this.getContractState(id);
     return state?.balances[wallet.toString()] ?? 0;
   }
+  /**
+   *
+   * @param srcCodeTransactionId - ArweaveTransactionId of the source code used to deploy contract
+   * @param initialState - typeof ANTContractState or ArweaveTransactionID - depending on the stateType chosen, either the json initial state of the contract to deploy, or the on-chain states TXID
+   * @param stateType - type of 'TXID' | 'TAG' | 'DATA' - TXID indicates use of a prexisting state on-chain, TAG and DATA use json states locally. DATA puts the state in the data key of the transaction, TAG puts the state in the tag.
+   * @returns
+   */
```
dtfiedler, line 105: let's remove these for now - we can use a linter plugin to auto-generate these so they are consistent

src/services/arweave/WarpDataProvider.ts
```diff
@@ -79,4 +96,97 @@ export class WarpDataProvider implements SmartweaveDataProvider {
...
+    tags = [],
+  }: {
+    srcCodeTransactionId: ArweaveTransactionID;
+    initialState: ANTContractJSON | ArweaveTransactionID;
+    stateType: 'TXID' | 'TAG' | 'DATA';
+    tags: TransactionTag[];
+  }): Promise<string | undefined> {
+    const stateSize = new Blob([JSON.stringify(initialState)]).size;
+    const tagSize = new Blob([JSON.stringify(tags)]).size;
+
```
dtfiedler, line 119: nit: use buffers (above)

src/types.ts
```diff
@@ -90,6 +90,17 @@ export interface SmartweaveDataProvider {
...
+    tags: TransactionTag[];
+  }): Promise<string | undefined>;
```
dtfiedler, line 103:
```suggestion
  }): Promise<string>;
```

src/services/arweave/WarpDataProvider.ts
atticusofsparta: I think we should: it leaves us open for supporting different schemas in the future for deployments, like if we want to extend the createAnt flow more. 

But to answer your question: no, its not currently necessary.

src/services/arweave/WarpDataProvider.ts
atticusofsparta: added

src/services/arweave/WarpDataProvider.ts
atticusofsparta: removed

atticusofsparta (approved):

atticusofsparta (approved):

PR #130, opened 2023-03-21
bug(fix):node is undefined

atticusofsparta:


dtfiedler (approved):
if it helps, sure. let's keep an eye on if it causes any unexpected issues.

PR #144, opened 2023-03-24
feat(PE-3320): setup sentry monitoring

dtfiedler:
## Details
Secrets have has been added as a GitHub secret on the repo. This value gets injected into the final build, hence why we construct it through various variables rather than just one. As an additional security layer, the Ingest has been scoped to only the GitHub domain.


arielmelendez (approved):
Consider assembling SENTRY_DSN_ENDPOINT inline rather than via variables.

PR #148, opened 2023-03-31
feat(PE-3302): search domain name dynamic url

dtfiedler:
## Details
Sets a `search` url param when searching for a name. This can later be extended to include details of registration like `Ant ID`, `Ticker`, `Years`, etc. that are all relevant for registering a name to create dynamic links for registering a name. In order to achieve proper `reset`, I moved `RegistrationProviderContext` to the top level, which would also allow us to register a name from anywhere in the app.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,8 +15,6 @@ import {
 import { About, Home, Manage, NotFound } from './components/pages';
 import { useArNSContract, useArweave } from './hooks/';
 import './index.css';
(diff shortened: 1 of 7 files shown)


dtfiedler:
there was one issue identified related to search resetting, i created a bug and will address in a separate ticket [PE-3395](https://ardrive.atlassian.net/browse/PE-3395)

atticusofsparta (approved):

PR #153, opened 2023-04-17
PE-3193: transaction page

atticusofsparta:
### Details
Leverages the `TransactionState` to manage the workflow for creating and posting transactions via the `/transaction` page. The intent of this state/workflow are to make posting transactions a consistent experience, and eventually enable dynamic linking. There was also some refactoring of types, to better use enums and consts.


src/hooks/useTransactionData/useTransactionData.tsx
```diff
@@ -0,0 +1,23 @@
...
+  transactionDataKeys.forEach((key: string) => {
+    const res = searchParams.get(key);
+    if (res) {
+      URLTransactionData[key] = res;
+    }
+  });
+  const URLContractType = searchParams.get('contractType');
+  const URLInteractionType = searchParams.get('interactionType');
+
+  // TODO: implement autofixes to type compatibility descrepencies ( if transaction data includes valid create ant keys for example, set the contract and interaction types appropriately)
```
matibat, line 18: Let me see if I get this: you want to set a type based on a value. Can you explain where/how exactly would this type be used?

I'm leaving the syntax for custom type wards which may be useful here: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

matibat (commented):
It's looking good! Left some comments mostly regarding string literals that can be constants

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -0,0 +1,415 @@
...
+import DeployTransaction from '../DeployTransaction/DeployTransaction';
+import TransactionComplete from '../TransactionComplete/TransactionComplete';
+import Workflow, { WorkflowStage } from '../Workflow/Workflow';
+
+function TransactionWorkflow({
+  contractType,
+  interactionType,
+  transactionData,
+  workflowStage,
+}: {
```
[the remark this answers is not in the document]
atticusofsparta: Unless there are multiple components (or functions / objects) that use the type, I prefer to right it directly in the component as its more readable.

However i would agree, better to have a named type if the type is used mutiple times.

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -0,0 +1,415 @@
...
+              return;
+            }
+            case 'successful': {
+              // completed section, no next
+              return;
+            }
+            case 'failed': {
+              // completed section, no next
+              return;
+            }
```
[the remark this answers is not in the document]
atticusofsparta: removed

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -0,0 +1,415 @@
...
+            default:
+              throw new Error(`Invalid workflow stage (${workflowStage})`);
+          }
+        default:
+          throw new Error(
+            `Invalid direction {${direction}}, Only next or back may be provided as directions`,
+          );
+      }
+    } catch (error) {
```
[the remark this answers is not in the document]
atticusofsparta: Actually, if the transaction data is bad, errors from that will bubble up and will be caught and logged here, so its still needed/good idea to catch here. Still need to setup the error display, but thats where the error will be sent eventually.

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -0,0 +1,415 @@
...
+function TransactionWorkflow({
+  contractType,
+  interactionType,
+  transactionData,
+  workflowStage,
+}: {
+  contractType: ContractType;
+  interactionType: AntInteraction | RegistryInteraction;
+  transactionData: Partial<TransactionData>;
+  workflowStage: 'pending' | 'confirmed' | 'successful' | 'failed';
```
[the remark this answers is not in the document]
karlprieb: nice "trick" I keep forgetting that typescript have this haha

src/components/layout/progress/StepProgressBar/StepProgressBar.tsx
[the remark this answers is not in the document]
atticusofsparta: yeah thats a good point. Im kind of dreading when we have to go back and implement theme changing.

src/App.tsx
```diff
@@ -76,7 +78,9 @@ function App() {
           path="transaction"
           element={
             <ProtectedRoute>
-              <Transaction />
+              <TransactionStateProvider reducer={transactionReducer}>
```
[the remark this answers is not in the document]
atticusofsparta: ah yes, removed. The intention of this was to have the state provider handling url data pulling, and to do so it had to be in the router context - BUT, i decided to move that logic to a different comp.

 Moved the state provider to top level.

src/components/layout/TransactionComplete/TransactionComplete.tsx
```diff
@@ -18,29 +18,30 @@ export enum transaction_types {
...
     <>
       <div className="flex-column center" style={{ gap: '3em' }}>
         <div className="flex-column center" style={{ gap: '2em' }}>
+          {/* TODO: configure error or fail states */}
           <AntCard
             domain={''}
             showTier={false}
             id={transactionId}
-            state={state}
+            state={state ?? undefined}
```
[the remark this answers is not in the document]
atticusofsparta: fixeded

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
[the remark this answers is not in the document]
atticusofsparta: used enum in this case since that matches current pattern.

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -0,0 +1,415 @@
...
+      contractType,
+      interactionType,
+    });
+    const newStages = getStagesByTransactionType({
+      contractType,
+      interactionType,
+    });
+    if (newSteps) setSteps(newSteps);
+    if (newStages) setStages(newStages);
+    if (workflowStage === 'confirmed' && deployedTransactionId) {
```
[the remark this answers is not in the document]
atticusofsparta: added

src/hooks/useTransactionData/useTransactionData.tsx
atticusofsparta: contract have individual functions - we have the REGISTRATION_INTERACTION enum and likewise for the ant interactions. We then have the RegistrationInteraction and AntInteraction types `(typeof REGISTRY_INTERACTION_TYPES)[keyof typeof REGISTRY_INTERACTION_TYPES];`. 

So - if the hook pulls a contract type of 'registry' but an interaction type of AntInteraction then that would be an incompatible contract-interaction type - even if functionName is named the same (because the props on that function name are probably different). 

So the validation would compare those values. Ideally the `functionName` from transaction data as well.

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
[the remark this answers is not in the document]
matibat: Yep. That's fine

src/hooks/useTransactionData/useTransactionData.tsx
matibat: I think this is what we talked about in the huddle 👍

src/App.tsx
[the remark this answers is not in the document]
dtfiedler: i see two approaches:
1. as is, and all the variables are required to be passed via environment variables from a given route, then navigate to the that route using `navigate` 
2. wrapping the full app in the transaction provider, dispatching changes for a transaction before navigating to the `/transaction` page, and loading the this state in the transaction page. you can then extend the `transaction` component to update the URL accordingly (and update the state if the URL provided is different the the transaction provider).

so, my question is did you consider both, and any reason that lead you to this implementation vs the other?

src/components/layout/TransactionComplete/TransactionComplete.tsx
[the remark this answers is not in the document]
dtfiedler: i don't see the commit that resolves this

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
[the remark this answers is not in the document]
dtfiedler: this is an eslint rule we can add for all files

src/main.tsx
```diff
@@ -5,16 +5,20 @@ import App from './App';
...
   <React.StrictMode>
     <GlobalStateProvider reducer={reducer}>
-      <RegistrationStateProvider reducer={registrationReducer}>
-        <App />
-      </RegistrationStateProvider>
+      <TransactionStateProvider reducer={transactionReducer}>
+        <RegistrationStateProvider reducer={registrationReducer}>
+          <App />
+        </RegistrationStateProvider>
+      </TransactionStateProvider>
```
dtfiedler, line 21: see comment related to approaches we can take - imo, i feel `TransactionStateProvider` should be available globally in the app and not just around the `Transaction` page.

atticusofsparta: agreed, moved it there

src/types.ts
```diff
@@ -312,21 +312,40 @@ export type AntInteraction =
...
+  ExtendLeasePayload &
+  UpgradeTierPayload &
+  TransferIOPayload &
+  SetTickerPayload &
+  SetControllerPayload &
+  SetNamePayload &
+  SetRecordPayload &
+  RemoveRecordPayload &
+  TransferAntPayload &
+  CreateAntPayload;
```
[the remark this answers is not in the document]
atticusofsparta: ahm, yes, im using this as a partial type in implementation `Partial<TransactionData>`

src/App.tsx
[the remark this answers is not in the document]
atticusofsparta: I made a hook that parses the url data into a consumable format for the transaction state provider, and the transaction page uses that to update the state provider as-needed. The hook also implements some validations for the data, and uses a function to get payload keys like contract function names.

The payload data (eg the data that is used in the contract interaction) is pulled as an array using the searchParams.getAll() function, and is then looped thru based on the payload keys for that interaction, which is retrieved from TRANSACTION_DATA_KEYS using the contractType and interactionType from the url.

src/types.ts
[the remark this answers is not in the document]
dtfiedler: `Partial` is a bit hacky in this context, as it basically reverts the type to be `any`. I rec using union types instead.

src/components/layout/TransactionComplete/TransactionComplete.tsx
```diff
@@ -18,37 +18,28 @@ export enum transaction_types {
   BUY_ARNS_NAME = 'Buy ArNS Name',
 }
 
-function TransactionSuccess({
+function TransactionComplete({
   transactionId = new ArweaveTransactionID(STUB_ARWEAVE_TXID),
-  state,
+  state, // eslint-disable-line
```
dtfiedler, line 23: why?

src/components/modals/CreateAntModal/CreateAntModal.tsx
```diff
@@ -258,7 +258,7 @@ function CreateAntModal() {
           <CloseIcon width={30} height={30} fill={'var(--text-white'} />
         </button>
         <Workflow
-          stage={workflowStage}
+          stage={workflowStage.toString()}
```
dtfiedler, line 261: why do we need `toString()` here?

src/state/contexts/TransactionState.tsx
```diff
@@ -1,21 +1,28 @@
...
 
 export type TransactionState = {
-  transactionData: TransactionData; // data that will be used to perform the transaction.
+  deployedTransactionId?: ArweaveTransactionID;
+  transactionData: Partial<TransactionData>; // data that will be used to perform the transaction.
   contractType: ContractType;
   interactionType: AntInteraction | RegistryInteraction;
-  workflowStage: 'pending' | 'confirmed' | 'deployed' | 'successful' | 'failed';
+  workflowStage: TRANSACTION_WORKFLOW_STATUS;
+  error?: any;
```
dtfiedler, line 25: nit: not needed - emit any errors to the `eventEmitter` and we'll handle errors globally

src/types.ts
```diff
@@ -307,68 +307,118 @@ export enum ANT_INTERACTION_TYPES {
...
 
 export const TRANSACTION_DATA_KEYS: {
   // specifying interaction types to the correct contract type, to ensure clarity and to prevent crossover of interaction types
-  [CONTRACT_TYPES.ANT]: { [K in AntInteraction]: string[] };
-  [CONTRACT_TYPES.REGISTRY]: { [K in RegistryInteraction]: string[] };
+  [CONTRACT_TYPES.ANT]: {
+    [K in AntInteraction]: { functionName: string; keys: string[] };
+  };
+  [CONTRACT_TYPES.REGISTRY]: {
+    [K in RegistryInteraction]: { functionName: string; keys: string[] };
```
dtfiedler, line 360: nit: `keys` might be better expressed as `requiredKeys` - also, these are duplicate types, so can abstract into one type

src/components/layout/TransactionComplete/TransactionComplete.tsx
atticusofsparta: oh. Wierd, thought I had to remove state, was having issue with ant card, removing something else must have fixed it.

src/state/contexts/TransactionState.tsx
atticusofsparta: removed

src/types.ts
atticusofsparta: I would agree, but not all of these keys are actually required per contract code - adding the type though

src/components/modals/CreateAntModal/CreateAntModal.tsx
atticusofsparta: changed the stage prop to a string type instead of number, it was complaining about the number - the workflow will be updated later though/

src/types.ts
dtfiedler: i mean:

```
export type InteractionInput = {
 functionName: string;
 keys: string[]
```
and using it as a type for lines 357 and 360

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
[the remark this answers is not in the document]
matibat: Ohh I think I created some confusion here: I wasn't saying we have to delete those but as these cases have the same logic we can consolidate them into the same thing:

```ts
switch(value) {
  case 1:
    print('it is one');
    break;
  case: 2:
    print('it is two');
    break;

  // clusters many cases corresponding to the same logic
  case 3:
  case 4:
  case 5:
     print('it is either 3, 4 or 5');
     break;

  default:
    print('it is not any of 1, 2, 3, 4, 5, or 6');
    break;
}
```

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -0,0 +1,391 @@
...
+              };
+            }
+            default:
+              throw new Error(
+                `Invalid workflow stage (${workflowStage}), workflow stage must be a number between 1 and 4`,
+              );
+          }
+        default:
+          throw new Error(
+            `Invalid TRANSACTIO TYPE {${contractType}}, Only registry or ant types may be provided as a transaction type`,
```
matibat, line 231: Nit
```suggestion
            `Invalid TRANSACTION TYPE {${contractType}}, Only registry or ant types may be provided as a transaction type`,
```

src/components/pages/Home/Home.tsx
```diff
@@ -75,8 +88,17 @@ function Home() {
       ) : (
         <>
           <Workflow
-            stage={stage}
+            stage={stage.toString()}
             onNext={() => {
+              if (stage == 1) {
+                const searchParameters = createSearchParams({
+                  data: [domain!, antID!.toString(), '1', '1'],
```
dtfiedler, line 95: this likely doesn't work - you cannot set an array of data in the URL, they'll all have a key of `data` so it's impossible to parse

dtfiedler: resolved in recent commits

src/components/pages/Transaction/Transaction.tsx
```diff
@@ -1,12 +1,24 @@
...
   const { transactionData, contractType, interactionType, workflowStage } =
     useTransactionData();
   const from = useLocation().state;
+  const [{}, dispatchTransactionState] = useTransactionState(); // eslint-disable-line
+
+  useEffect(() => {
+    dispatchTransactionState({
+      type: 'setWorkflowStage',
+      payload: TRANSACTION_WORKFLOW_STATUS.PENDING,
+    });
```
dtfiedler, line 20: what about `reset` at the end of `TransactionWorkflow` instead of force updating here?

src/components/layout/progress/StepProgressBar/StepProgressBar.tsx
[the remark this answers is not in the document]
dtfiedler: agree - individual styles attributes on components is brittle. we avoided using a style system/framework early on (sass, tailwind, etc.) to avoid a steep learning curve. but a v2 of this app might want to consider those tools

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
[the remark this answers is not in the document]
dtfiedler: ah this was the pattern is was trying to sniff out @matibat! great example

src/components/pages/Transaction/Transaction.tsx
atticusofsparta: if we reset the workflow stage it will rerender us back to the pending stage..... that needs to be done either at the beginning or in the state provider.... even in the state provider im not sure that works, i think it has to always be at the beginning. eg when the transaction page loads.

dtfiedler: hmm, this might cause issues when we do add in dynamic URL loading - as it'll ignore what `useTransactionData` hook will parse and set in the provider state

dtfiedler: for now, let's do this. we'll confirm that the page was loaded from an internal page when setting this. that way, any internal routing will start fresh, and when we add dynamic linking, `from` will be empty, so it won't override the state set by `useTransactionData`

```
  useEffect(() => {
    if (from) {
      dispatchTransactionState({
        type: 'setWorkflowStage',
        payload: TRANSACTION_WORKFLOW_STATUS.PENDING,
      });
    }
  }, [from]);
```

dtfiedler: fine as is for now, just wanted to call this out.

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -0,0 +1,406 @@
...
+              setSteps(newSteps);
+              dispatchTransactionState({
+                type: 'setWorkflowStage',
+                payload: TRANSACTION_WORKFLOW_STATUS.CONFIRMED,
+              });
+              const originalTxId = await arweaveDataProvider.writeTransaction(
+                new ArweaveTransactionID(assetId),
+                {
+                  function: functionName,
+                  payload,
```
dtfiedler, line 106: https://viewblock.io/arweave/tx/wcAQhzSnJ3OoA6QzLt9AihFx2p2uH5luiWDGC8mdFII

the payload needs to be spread

```suggestion
                  ...payload,
```

src/components/layout/TransactionComplete/TransactionComplete.tsx
```diff
@@ -18,37 +18,29 @@ export enum transaction_types {
...
-            }}
-            disabledKeys={[
-              'tier',
-              'evolve',
-              'maxSubdomains',
-              'domain',
-              'leaseDuration',
-            ]}
+            state={state}
           />
```
dtfiedler, line 44: image

after successfully deploying a name, I'm not seeing what I would expect for `domain` and `leaseDuration`

src/components/layout/UpgradeTier/UpgradeTier.tsx
```diff
@@ -9,7 +9,7 @@ import {
   SMARTWEAVE_TAG_SIZE,
   TIER_DATA,
 } from '../../../utils/constants';
-import { calculateArNSNamePrice } from '../../../utils/searchUtils';
+import { calculateArNSNamePrice } from '../../../utils/searchUtils/searchUtils';
```
dtfiedler, line 12: you can reference just `/utils` now

src/components/pages/Home/Home.tsx
```diff
@@ -19,7 +19,7 @@ import {
 import {
   isArNSDomainNameAvailable,
   isArNSDomainNameValid,
-} from '../../../utils/searchUtils';
+} from '../../../utils/searchUtils/searchUtils';
```
dtfiedler, line 22: nit: reference just `/utils`

src/services/arweave/SimpleArweaveDataProvider.ts
```diff
@@ -8,11 +8,11 @@ import {
   TransactionHeaders,
   TransactionTag,
 } from '../../types';
+import { tagsToObject } from '../../utils';
```
dtfiedler, line 11: ❤️

src/components/layout/TransactionComplete/TransactionComplete.tsx
```diff
@@ -1,55 +1,49 @@
...
+  ArweaveTransactionID,
+  ContractType,
+  RegistryInteraction,
+  TransactionData,
+} from '../../../types';
 import {
   STUB_ARWEAVE_TXID,
   WARP_CONTRACT_BASE_URL,
 } from '../../../utils/constants';
+import { getArNSMappingByInteractionType } from '../../../utils/transactionUtils/transactionUtils';
```
dtfiedler, line 17: nit: use `utils`

dtfiedler (approved):
looks good!

PR #154, opened 2023-04-19
feat(PE-3493) - use pdns cache, remove arweaveDataProvider from global state, cleanup types

dtfiedler:
### Details
- Removes `arweaveDataProvider` from global state and is now provided via `useArweaveCompositeDataProvider`
- updates `getContractState()` to accept a type, to allow the function to return `ArNSContractJSON` and `ANTContractJSON` for better type checking
- uses `AntContract` class to make retrieving values safer/more consistent and with types


src/services/arweave/ArweaveCompositeDataProvider.ts
```diff
@@ -1,36 +1,46 @@
...
     return this._arweaveProvider.getWalletBalance(id);
   }
 
-  async getContractState(
+  async getContractState<T extends ArNSContractJSON | ANTContractJSON>(
     id: ArweaveTransactionID,
-  ): Promise<ArNSContractJSON | ANTContractJSON | undefined> {
-    return this._warpProvider.getContractState(id);
+  ): Promise<T> {
+    return Promise.any(
```
atticusofsparta, line 41: why `any` vs `race` on the promise?

dtfiedler: we want it to fallback to `warp` if we get a bad response from the pdns-service. `any` gives us the first `resolved` promise, whereas `race` returns the first `rejection` or `resolve`https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race#return_value

atticusofsparta (approved):

PR #155, opened 2023-04-20
PE-3193: determine the enum value based off of its constant values instead of casting the type

matibat:
This is a suggestion to open a discussion about dynamic values being mapped to (constant) enum values.

My main concern here: we were previously casting a string into an enum. And:
1. When utilizing the keyword `as`, we're telling TypeScript that it can trust that the value has a certain type.
2. After compiled, the casts will be removed. E.i. casting a value won't have any effect on the logic.
3. When the enum is evaluated in a switch/case, it doesn't make sense to have the default case if we cover every possible value the enum can have.

So what I'd like from this PR is to treat unknown/unexpected values accordingly by either throwing an error when we parse the dynamic value (which comes from the search parameters) or by having a placeholder value for the unknown case (see the UNKNOWN values I added to the enums).

### DIFF
--- a/src/hooks/useTransactionData/useTransactionData.tsx
+++ b/src/hooks/useTransactionData/useTransactionData.tsx
@@ -1,7 +1,14 @@
 import { useEffect, useState } from 'react';
 import { useSearchParams } from 'react-router-dom';
 
-import { AntInteraction, ContractType, RegistryInteraction } from '../../types';
+import {
+  ANT_INTERACTION_TYPES,
+  AntInteraction,
+  CONTRACT_TYPES,
+  ContractType,
+  REGISTRY_INTERACTION_TYPES,
+  RegistryInteraction,
+} from '../../types';
 import { getTransactionPayloadByInteractionType } from '../../utils/searchUtils';
 
 function useTransactionData() {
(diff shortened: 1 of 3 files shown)


src/hooks/useTransactionData/useTransactionData.tsx
```diff
@@ -12,18 +19,34 @@ function useTransactionData() {
   const [URLAssetId, setURLAssetId] = useState<string>(
     () => searchParams.get('assetId') ?? '',
   );
-  const [URLContractType, setURLContractType] = useState<ContractType>(
-    () =>
-      (searchParams.get('contractType') as ContractType) ??
-      ('' as ContractType),
-  );
+  const [URLContractType, setURLContractType] = useState<ContractType>(() => {
```
matibat, line 22: Removes the cast for `contractType`

src/hooks/useTransactionData/useTransactionData.tsx
```diff
@@ -12,18 +19,34 @@ function useTransactionData() {
...
+    }
+  });
   const [URLInteractionType, setURLInteractionType] = useState<
     AntInteraction | RegistryInteraction
-  >(
-    (searchParams.get('interactionType') as
-      | AntInteraction
-      | RegistryInteraction) ?? ('' as AntInteraction | RegistryInteraction),
-  );
+  >(() => {
```
matibat, line 34: Removes the cast for `interactionType`

src/types.ts
matibat, line 1: Adds the UNKNOWN value to each enum

src/utils/searchUtils.ts
matibat, line 1: Safewards to avoid throw if we get the UNKNOWN value here

matibat (commented):
Explanation of the changes

atticusofsparta (approved):

src/utils/searchUtils.ts
matibat: I think this part can be avoided by having a type that already excludes UNKNOWN... Let me think...

PR #163, opened 2023-04-21
Feat(PE-3361): edit name

atticusofsparta:


src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -229,3 +255,13 @@ export function getArNSMappingByInteractionType(
...
+
+export function getInteractionTypeFromField(field: string) {
+  switch (field) {
+    case 'name':
+      return ANT_INTERACTION_TYPES.SET_NAME;
+
+    default:
+      return undefined;
+  }
+}
```
dtfiedler, line 267: consider using an object for these mappings. 

e.g.
```
export const FieldToInteractionMap = {
    'name': ANT_INTERACTION_TYPES.NAME
}
```
and you can access the object directly or update the function
```
export function  getInteractionTypeFromField(field: string): {
    // maybe check it exists and throw error if it doesn't
    return FieldToInteractionMap[field]
}
```

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -203,24 +208,45 @@ export function getArNSMappingByInteractionType(
       break;
     case CONTRACT_TYPES.ANT: {
       switch (interactionType) {
-        case ANT_INTERACTION_TYPES.CREATE: {
-          const data = transactionData as CreateAntPayload;
+        case ANT_INTERACTION_TYPES.CREATE:
+          {
+            const data = transactionData as CreateAntPayload;
```
dtfiedler, line 213: you could us the `isObjectOfTransactionPayloadType<CreateAntPayload>` function here so you wouldn't even have to cast.

atticusofsparta: added

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -210,58 +210,78 @@ export function getArNSMappingByInteractionType(
...
+                  'evolve',
+                  'maxSubdomains',
+                  'id',
+                  'domain',
+                  'leaseDuration',
+                ],
+              };
+            }
+          }
+          break;
```
dtfiedler, line 243: what happens in the failed if case? i would consider throwing an error

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -99,9 +100,9 @@ export function getTransactionPayloadByInteractionType(
   contractType: ContractType,
   interactionType: AntInteraction | RegistryInteraction,
   data: string | string[],
-) {
+): TransactionData | undefined {
```
dtfiedler, line 103: nit: why undefined?

src/components/modals/ManageAntModal/ManageAntModal.tsx
```diff
@@ -248,10 +256,33 @@ function ManageAntModal() {
...
-                          );
-                          // TODO: write contract interaction
+                          const payload =
+                            getTransactionPayloadByInteractionType(
+                              CONTRACT_TYPES.ANT,
+                              row.interactionType,
+                              modifiedValue!,
+                            );
+
+                          if (payload && row.interactionType) {
```
dtfiedler, line 266: add `id` check here, don't force cast below

PR #165, opened 2023-04-24
feat(PE-3360): added edit ticker config

atticusofsparta:
transaction example: https://viewblock.io/arweave/tx/AZx7E4bJqivL2kFuyKzu7_3bEpsYkSC3qm9kqST3oY0


dtfiedler:
rebase with develop

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -185,6 +187,7 @@ export function getArNSMappingByInteractionType(
     transactionData: TransactionData;
   },
 ): ArNSMapping | undefined {
+  let mapping: ArNSMapping = { domain: '' };
```
dtfiedler, line 190: return within the case

atticusofsparta: added

dtfiedler:
nit: add the ticket number to the github title so JIRA pulls it in

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -264,6 +264,7 @@ function TransactionWorkflow({
       switch (contractType) {
         case CONTRACT_TYPES.ANT: {
           switch (interactionType) {
+            case ANT_INTERACTION_TYPES.SET_TICKER:
```
dtfiedler, line 267: interesting we're using a switch here but likely man of these will fall through to do the same thing. let's consider just doing an if check. switches within switches is tough to maintain and comprehend

dtfiedler:
please update the description to include some example TX interactions related to changing a ticker name

dtfiedler (approved):

PR #166, opened 2023-04-24
feat(PE-3363): edit targetID config

atticusofsparta:
example transaction: https://viewblock.io/arweave/tx/btNwLXV9aqtja_8R3tRcD2DBVTV9Iyn4sPC-O76kX_o
image

### DIFF
--- a/src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
+++ b/src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
@@ -228,6 +228,13 @@ function TransactionWorkflow({
                 3: { title: 'Complete', status: '' },
               };
             }
+            case ANT_INTERACTION_TYPES.SET_TARGET_ID: {
+              return {
+                1: { title: 'Confirm Target ID', status: 'pending' },
+                2: { title: 'Deploy Target ID Change', status: '' },
+                3: { title: 'Complete', status: '' },
+              };
+            }
             default:
               throw new Error(
                 `Invalid workflow stage (${workflowStage}), workflow stage must be a number between 1 and 4`,
@@ -264,6 +271,7 @@ function TransactionWorkflow({
       switch (contractType) {
         case CONTRACT_TYPES.ANT: {
           switch (interactionType) {
+            case ANT_INTERACTION_TYPES.SET_TARGET_ID:
             case ANT_INTERACTION_TYPES.SET_TICKER:
             case ANT_INTERACTION_TYPES.SET_NAME: {
               return {
--- a/src/components/modals/ManageAntModal/ManageAntModal.tsx
+++ b/src/components/modals/ManageAntModal/ManageAntModal.tsx
@@ -9,6 +9,7 @@ import { useGlobalState } from '../../../state/contexts/GlobalState';
 import { useTransactionState } from '../../../state/contexts/TransactionState';
 import {
   ANTContractJSON,
+  ANT_INTERACTION_TYPES,
   ArNSRecordEntry,
   ArweaveTransactionID,
   CONTRACT_TYPES,
@@ -18,7 +19,7 @@ import {
 } from '../../../types';
 import {
   getInteractionTypeFromField,
-  getTransactionPayloadByInteractionType,
+  mapTransactionDataKeyToPayload,
 } from '../../../utils';
 import eventEmitter from '../../../utils/events';
 import { mapKeyToAttribute } from '../../cards/AntCard/AntCard';
@@ -54,6 +55,7 @@ function ManageAntModal() {
   const arweaveDataProvider = useArweaveCompositeProvider();
   const [{ arnsSourceContract }] = useGlobalState();
   const [{}, dispatchTransactionState] = useTransactionState(); // eslint-disable-line
+  const [antState, setAntState] = useState<ANTContract>();
   const [antName, setAntName] = useState<string>();
   const [editingField, setEditingField] = useState<string>();
   const [modifiedValue, setModifiedValue] = useState<string | number>();
@@ -87,6 +89,7 @@ function ManageAntModal() {
       arweaveDataProvider.getTransactionStatus(txId),
     ]);
     const ant = new ANTContract(contractState);
+    setAntState(ant);
     const record = Object.values(arnsSourceContract.records).find(
       (r) => r.contractTxId === txId.toString(),
     );
@@ -303,12 +306,27 @@ function ManageAntModal() {
                             borderColor: 'var(--accent)',
                           }}
                           onClick={() => {
+                            // TODO update when ant source code is updated to not overwrite existing info
                             const payload =
-                              getTransactionPayloadByInteractionType(
-                                CONTRACT_TYPES.ANT,
-                                row.interactionType,
-                                modifiedValue?.toString(),
-                              );
+                              row.interactionType ===
+                              ANT_INTERACTION_TYPES.SET_TARGET_ID
+                                ? mapTransactionDataKeyToPayload(
+                                    CONTRACT_TYPES.ANT,
+                                    row.interactionType,
+                                    [
+                                      '@',
+                                      modifiedValue!.toString(),
+                                      antState!.records[
+                                        '@'
+                                      ].ttlSeconds.toString(),
+                                    ],
+                                  )
+                                : mapTransactionDataKeyToPayload(
+                                    CONTRACT_TYPES.ANT,
+                                    row.interactionType,
+                                    modifiedValue!.toString(),
+                                  );
+
                             if (payload && row.interactionType && id) {
                               // eslint-disable-next-line
                               const { assetId, functionName, ...data } =
--- a/src/components/modals/ManageAntModal/ManageAntModal.tsx
+++ b/src/components/modals/ManageAntModal/ManageAntModal.tsx
@@ -14,6 +14,7 @@ import {
   CONTRACT_TYPES,
   ManageAntRow,
   TRANSACTION_TYPES,
+  VALIDATION_INPUT_TYPES,
 } from '../../../types';
 import {
   getInteractionTypeFromField,
@@ -22,6 +23,7 @@ import {
 import eventEmitter from '../../../utils/events';
 import { mapKeyToAttribute } from '../../cards/AntCard/AntCard';
 import { ArrowLeft, CloseIcon, PencilIcon } from '../../icons';
+import ValidationInput from '../../inputs/text/ValidationInput/ValidationInput';
 import TransactionStatus from '../../layout/TransactionStatus/TransactionStatus';
 
 const EDITABLE_FIELDS = [
@@ -54,7 +56,7 @@ function ManageAntModal() {
   const [{}, dispatchTransactionState] = useTransactionState(); // eslint-disable-line
   const [antName, setAntName] = useState<string>();
   const [editingField, setEditingField] = useState<string>();
-  const [modifiedValue, setModifiedValue] = useState<string>();
+  const [modifiedValue, setModifiedValue] = useState<string | number>();
   const [rows, setRows] = useState<ManageAntRow[]>([]);
 
   useEffect(() => {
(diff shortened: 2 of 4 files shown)


src/components/modals/ManageAntModal/ManageAntModal.tsx
```diff
@@ -268,7 +307,7 @@ function ManageAntModal() {
                               getTransactionPayloadByInteractionType(
                                 CONTRACT_TYPES.ANT,
                                 row.interactionType,
-                                modifiedValue!,
+                                modifiedValue!.toString(),
```
dtfiedler, line 310: remove the force case (`!`) here and update the function to handle a potentially `undefined` for this param

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -125,18 +126,28 @@ export function getTransactionPayloadByInteractionType(
...
+                }
+                payload[key] = txData[index];
+              },
+            );
+          }
+          break;
+        case ANT_INTERACTION_TYPES.SET_TARGET_ID:
+          {
+            payload[
+              TRANSACTION_DATA_KEYS[contractType][interactionType].keys[0]
```
dtfiedler, line 146: why do we need different logic here?

atticusofsparta: good point, no reason not to do the same logic. Still needs the extra subdomain setting though. Thats there so we dont have to set it in jsx on the tables.

atticusofsparta: added

src/components/modals/ManageAntModal/ManageAntModal.tsx
atticusofsparta: added

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -125,18 +131,35 @@ export function getTransactionPayloadByInteractionType(
...
+              (key: string, index) => {
+                if (!txData[index]) {
+                  throw new Error(
+                    `Missing key (${key}) from transaction data in the url. This may be due to the order of the data, the current order is [${txData}], the correct order is [${TRANSACTION_DATA_KEYS[contractType][interactionType].keys}]`,
+                  );
+                }
+                payload[key] = txData[index];
+              },
+            );
+            payload.subDomain = '@';
```
dtfiedler, line 160: move this to a function

atticusofsparta: moveded

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -326,10 +320,84 @@ export const FieldToInteractionMap: {
...
+            );
+          }
+          payload[key] = data[index];
+        },
+      );
+    }
+  } catch (error) {
+    console.log(error);
+  } finally {
+    return payload; // eslint-disable-line
```
dtfiedler, line 401: why is eslint disabled? and what happens if this is an empty object?

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -101,72 +102,32 @@ export function isInteractionCompatible({
...
-      TRANSACTION_DATA_KEYS[contractType][interactionType].keys.forEach(
-        (key, index) => {
-          if (!txData[index]) {
-            // if missing transaction data from the url, throw an error
-            throw new Error(
-              `Missing key (${key}) from transaction data in the url. This may be due to the order of the data, the current order is [${txData}], the correct order is [${TRANSACTION_DATA_KEYS[contractType][interactionType].keys}]`,
-            );
+        case ANT_INTERACTION_TYPES.SET_TARGET_ID:
+          {
+            payload.subDomain = '@';
```
dtfiedler, line 127: so just one case in the switch?

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -264,6 +271,7 @@ function TransactionWorkflow({
       switch (contractType) {
         case CONTRACT_TYPES.ANT: {
           switch (interactionType) {
+            case ANT_INTERACTION_TYPES.SET_TARGET_ID:
             case ANT_INTERACTION_TYPES.SET_TICKER:
             case ANT_INTERACTION_TYPES.SET_NAME: {
               return {
```
dtfiedler, line 277: still don't see the point of switch here

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -326,10 +320,84 @@ export const FieldToInteractionMap: {
...
+                TRANSACTION_DATA_KEYS[CONTRACT_TYPES.ANT][interactionType].keys
+              }]`,
+            );
+          }
+          payload[key] = data[index];
+        },
+      );
+    }
+  } catch (error) {
+    console.log(error);
```
dtfiedler, line 399: we're swallowing the error here. when it comes to errors we need to either
1. throw them and let the caller decide how to handle
2. emit them to the eventEmitter so they get captured by sentry and displayed

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
atticusofsparta: type protection for accessing TRANSACTION_DATA_KEYS

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: yes, the switch is now for overrides on special interaction types

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: odd, not needed i guess.

dtfiedler:
keep in mind, a switch statement can be swapped for a defined object or `Map`, which are generally easier to read, maintain and extend. the level of deeply nested switch statements we have concerns me. not a blocker, just something to think about

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
dtfiedler: > type protection for accessing TRANSACTION_DATA_KEYS
not sure i follow, can you explain what you mean?

atticusofsparta: If we dont use a switch, or two `if` statements checking the type in the interaction, where we access the TRANSACTION_DATA_KEYS

src/utils/transactionUtils/transactionUtils.ts
dtfiedler: i'm of the opinion all of these util functions should throw, and events are emitted from calling `.tsx` components`.

atticusofsparta: ah ok, makes sense, fixing

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -326,10 +330,85 @@ export const FieldToInteractionMap: {
...
+          }
+          payload[key] = data[index];
+        },
+      );
+      return payload;
+    }
+  } catch (error: any) {
+    console.log(error);
+    throw new Error(error);
+  }
```
dtfiedler, line 413: just remove the `try/catch` and emit the error in the component

atticusofsparta: fixed

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -326,10 +330,80 @@ export const FieldToInteractionMap: {
...
+        // if missing transaction data from the url, throw an error
+        throw new Error(
+          `Missing key (${key}) from transaction data in the url. This may be due to the order of the data, the current order is [${data}], the correct order is [${
+            TRANSACTION_DATA_KEYS[CONTRACT_TYPES.REGISTRY][interactionType].keys
+          }]`,
+        );
+      }
+      payload[key] = data[index];
+    });
+    return payload;
```
dtfiedler, line 378: you _could_ use a reducer here to do this and just return an object instead of appending `payload` object. but fine for now.

dtfiedler:
when testing:
```
Error: transaction data not of correct payload type <SetRecordPayload>
    at getTransactionPayloadByInteractionType (transactionUtils.ts:135:21)
    at onClick (ManageAntModal.tsx:307:31)
    at HTMLUnknownElement.callCallback2 (react-dom.development.js:4164:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:16)
    at invokeGuardedCallback (react-dom.development.js:4277:31)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4291:25)
    at executeDispatch (react-dom.development.js:9041:3)
    at processDispatchQueueItemsInOrder (react-dom.development.js:9073:7)
    at processDispatchQueue (react-dom.development.js:9086:5)
    at dispatchEventsForPlugins (react-dom.development.js:9097:3)
o
```

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -101,72 +102,44 @@ export function isInteractionCompatible({
...
-          if (!txData[index]) {
-            // if missing transaction data from the url, throw an error
-            throw new Error(
-              `Missing key (${key}) from transaction data in the url. This may be due to the order of the data, the current order is [${txData}], the correct order is [${TRANSACTION_DATA_KEYS[contractType][interactionType].keys}]`,
-            );
+        case ANT_INTERACTION_TYPES.SET_TARGET_ID:
+          {
+            if (
+              !isObjectOfTransactionPayloadType<SetRecordPayload>(
+                payload!,
```
dtfiedler, line 129: don't force cast - add to the if check above

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -101,72 +102,44 @@ export function isInteractionCompatible({
...
+                TRANSACTION_DATA_KEYS[CONTRACT_TYPES.ANT][
+                  ANT_INTERACTION_TYPES.SET_RECORD
```
dtfiedler, line 131: shouldn't this be `SET_TARGET_ID`?

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -326,10 +330,80 @@ export const FieldToInteractionMap: {
...
+      if (!data[index]) {
+        // if missing transaction data from the url, throw an error
+        throw new Error(
+          `Missing key (${key}) from transaction data in the url. This may be due to the order of the data, the current order is [${data}], the correct order is [${
+            TRANSACTION_DATA_KEYS[CONTRACT_TYPES.REGISTRY][interactionType].keys
+          }]`,
+        );
+      }
+      payload[key] = data[index];
+    });
```
dtfiedler, line 377: i was hoping to see _this_ in it's own function that just takes the `CONTRACT_TYPE` AND `interactionType` to return the payload with all the keys. it's the exact same logic below 😢

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: set target id is a subtype of set record, the actual function is SET_RECORD, or 'setRecord' is the function on the contract. SET_TARGET_ID is specifically deploying to the @ record. So, the payload is supposed to be SetRecordPayload.

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: due to the funky nature of the TRANSACTION_DATA_KEYS object, this is necesary to duplicated for every contract instance. Unless you know of a better way.

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: added overides to util to allow for this

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: fixed

src/utils/transactionUtils/transactionUtils.ts
dtfiedler: this makes me sad. i think there is opportunity to improve this so please add a `TODO` comment here. I'd imagine with some of @matibat 's type changes, we can more effectively implement this util.

atticusofsparta: added todo

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: removed overrides and made more payload types for sub-payload-types

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: added todo

dtfiedler:
submitted https://viewblock.io/arweave/tx/wNN6Aoi0IWRwM-3bUCetHQiHE3htQvjmc1_Et0gx9HQ

src/types.ts
```diff
@@ -290,9 +290,10 @@ export type SetNamePayload = {
 };
 export type SetRecordPayload = {
   subDomain: string;
-  transactionId: string;
-  ttlSeconds: number;
 };
+export type SetTargetIDPayload = {
+  transactionId: string;
+} & SetRecordPayload;
```
dtfiedler, line 296: nit: space between types

atticusofsparta: fixed

dtfiedler:
submitted one change:
image

to contract:
image

and here's the result:
image


so takeaways:
- we should not overwrite other fields when setting the ticker, we should submit `setRecord` with any existing fields on the record and only update `transactionId`

dtfiedler (approved):

dtfiedler:
please test heavily after it's deployed

PR #178, opened 2023-04-27
chore(PE-3577) : Update usage of ARNS and ANT to PDNS and PDNT

atticusofsparta:


src/index.css
```diff
@@ -543,21 +543,21 @@ input[type='text']:disabled {
...
 
-.ant-pagination-item > a {
+.pdnt-pagination-item > a {
   color: white !important;
 }
 
-.ant-pagination-item-active > a {
+.pdnt-pagination-item-active > a {
   background-color: var(--text-faded) !important;
 }
```
dtfiedler, line 563: don't switch these - they are classes from `antd`

dtfiedler (approved):

PR #179, opened 2023-04-27
feat(PE-3587): edit ttl seconds

atticusofsparta:


src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -235,6 +235,13 @@ function TransactionWorkflow({
...
               };
             }
+            case PDNT_INTERACTION_TYPES.SET_TTL_SECONDS: {
+              return {
+                1: { title: 'Confirm TTL Seconds', status: 'pending' },
+                2: { title: 'Deploy TTL Seconds Change', status: '' },
+                3: { title: 'Complete', status: '' },
+              };
+            }
             default:
```
dtfiedler, line 245: still my opinon we put this switch case into an object in a `constants` file

src/components/modals/ManagePDNTModal/ManagePDNTModal.tsx
```diff
@@ -319,6 +320,20 @@ function ManagePDNTModal() {
...
+                                  PDNT_INTERACTION_TYPES.SET_TTL_SECONDS
+                                ? mapTransactionDataKeyToPayload(
+                                    CONTRACT_TYPES.PDNT,
+                                    row.interactionType,
+                                    [
+                                      '@',
+                                      pdntState!.records['@'].transactionId
+                                        .length
+                                        ? pdntState!.records['@'].transactionId
+                                        : STUB_ARWEAVE_TXID,
```
dtfiedler, line 333: i'm not a fan of stubbing the tx id

atticusofsparta: niether am I but its required per contract code else we cannot se the ttl seconds 
image

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
atticusofsparta: moved this the transaction data object and created an accessor util

PR #182, opened 2023-04-28
fix(PE-3574): updated transaction complete links to route to correct pages, …

atticusofsparta:
…updated util to correctly copy steps object, updated create pdnt to use transaction workflow


src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -80,30 +78,40 @@ function TransactionWorkflow({
...
                   type: 'setDeployedTransactionId',
-                  payload: originalTxId,
+                  payload: new ArweaveTransactionID(originalTxId.toString()),
                 });
               }
-              newSteps['2'].status = 'success';
-              newSteps['3'].status = 'success';
-              setSteps(newSteps);
+              steps['2'].status = 'success';
+              steps['3'].status = 'success';
```
dtfiedler, line 114: nit: put this in a function and call it from the switch statement. it's hard to understand what's going on in here with it nested

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -174,7 +195,7 @@ export const getWorkflowStepsForInteraction = (
     status: string;
   };
 } => {
-  return WorkflowStepsForInteractions[interaction];
+  return _.cloneDeep(WorkflowStepsForInteractions[interaction]);
```
dtfiedler, line 198: `structuredClone()` is a built-in function for deep-copying. lodash is a large package, so avoid it if we can

https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -380,9 +390,9 @@ export function getInteractionTypeFromField(field: string) {
 
 export function mapTransactionDataKeyToPayload(
   interactionType: ValidInteractionType,
-  data: string | number | Array<string | number>,
+  data: string | number | Array<string | number | PDNTContractJSON>,
```
dtfiedler, line 393: i think we should just change this to an object vs. an array, but can be done in another PR.

src/utils/transactionUtils/transactionUtils.ts
```diff
@@ -405,3 +415,28 @@ export function mapTransactionDataKeyToPayload(
...
+  if (
+    interactionType === INTERACTION_TYPES.CREATE &&
+    isObjectOfTransactionPayloadType<CreatePDNTPayload>(
+      transactionData,
+      TRANSACTION_DATA_KEYS[INTERACTION_TYPES.CREATE].keys,
+    )
+  ) {
+    return transactionData.deployedTransactionId?.toString() ?? '';
+  }
+  return transactionData.assetId.toString() ?? '';
```
dtfiedler, line 441: i don't think `assetId` can be null, so no null-coalesce needed here

dtfiedler (approved):
functionally looks good - loving seeing `/create` fold into the `/transaction` workflow. some small nits to cleanup.

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: you may be right.... i was originally using this function for url params were all the data was stored in the url under the `data` search params, and was sticking with that logic here.

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
atticusofsparta: added

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: added

src/utils/transactionUtils/transactionUtils.ts
atticusofsparta: added

dtfiedler (approved):

PR #183, opened 2023-04-28
feat(PE-3362): edit controller

atticusofsparta:


dtfiedler:
i'm surprised not see any `ValidationInput's` for the `controller` field in the `ManagePDNTModal`

src/components/layout/TransactionWorkflow/TransactionWorkflow.tsx
```diff
@@ -166,6 +166,7 @@ function TransactionWorkflow({
       throw Error('Unable to get PDNT properties.');
     }
     switch (interactionType) {
+      case INTERACTION_TYPES.SET_CONTROLLER:
       case INTERACTION_TYPES.CREATE:
       case INTERACTION_TYPES.SET_TTL_SECONDS:
       case INTERACTION_TYPES.SET_TARGET_ID:
```
dtfiedler, line 172: we have the array of `antInteractionTypes` that we could use for these checks

PR #197, opened 2023-05-04
Feat(PE-3643): add create undername interaction

atticusofsparta:
image

note: currently the gateway is set to use ar-io.dev - this occasionally has issues on the gateway side with posting transactions, if so just change the gateway url in GlobalState.tsx to arweave.net, and it should work.

### DIFF
--- a/src/components/layout/Undernames/Undernames.tsx
+++ b/src/components/layout/Undernames/Undernames.tsx
@@ -71,6 +71,7 @@ function Undernames() {
   const [tablePage, setTablePage] = useState<number>(1);
 
   // modal state
+  const [confirmations, setConfirmations] = useState(0);
   const [action, setAction] = useState<
     UndernameTableInteractionTypes | undefined
   >();
@@ -88,6 +89,10 @@ function Undernames() {
     arweaveDataProvider
       .getContractState<PDNTContractJSON>(new ArweaveTransactionID(id))
       .then((state) => setPDNTState(state));
+
+    arweaveDataProvider
+      .getTransactionStatus(new ArweaveTransactionID(id))
+      .then((res) => setConfirmations(res));
   }, [id]);
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


dtfiedler:
rebase with develop so the diff shows only relevant changes

src/components/layout/Undernames/Undernames.tsx
```diff
@@ -81,12 +116,100 @@ function Undernames() {
...
+      const confirmations = await arweaveDataProvider.getTransactionStatus(
+        new ArweaveTransactionID(id),
+      );
+      if (confirmations < 1) {
+        throw new Error(
+          'PDNT must have a minimum of 1 confirmation before editing, this protects against editing contracts lost due to block reorgs.',
+        );
+      }
+      switch (action) {
+        case UNDERNAME_TABLE_ACTIONS.CREATE:
```
dtfiedler, line 143: i'm assuming this is a switch bc other actions logic will be added here?

atticusofsparta: yup, edit and remove

src/components/layout/Undernames/Undernames.tsx
```diff
@@ -112,6 +126,72 @@ function Undernames() {
...
   }
+  async function handleOnNext() {
+    try {
+      if (!id) {
+        throw new Error('No PDNT ID found, unable to perform transaction.');
+      }
+      const confirmations = await arweaveDataProvider.getTransactionStatus(
+        new ArweaveTransactionID(id),
+      );
+      if (confirmations < 1) {
```
dtfiedler, line 137: Should this be something like 15?

dtfiedler (commented):
I think we should set required confirmations to 15+ before editing. The table disables the manage button depending on the # of confirmations so we should match that.

src/components/layout/Undernames/Undernames.tsx
atticusofsparta: added

src/components/layout/Undernames/Undernames.tsx
```diff
@@ -112,6 +131,70 @@ function Undernames() {
     setAction(undefined);
     setSelectedRow(undefined);
   }
+  function handleOnNext() {
+    try {
+      if (!id) {
+        throw new Error('No PDNT ID found, unable to perform transaction.');
+      }
+
+      if (confirmations < 15) {
```
dtfiedler, line 140: nit: do we have a constant for this?

dtfiedler (approved):

PR #198, opened 2023-05-04
Pe 3646 convert table to antd

atticusofsparta:


src/index.css
```diff
@@ -440,6 +440,109 @@ input[type='text']:disabled {
   height: 50px;
 }
 
+.ant-table-content > table {
+  border-spacing: 0px 10px;
+  width: 100%;
+  background-color: var(--bg-color);
+  border: none !important;
+}
+.ant-table {
```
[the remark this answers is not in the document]
atticusofsparta: uhm.... how?

src/index.css
```diff
@@ -440,6 +440,109 @@ input[type='text']:disabled {
...
+  border-radius: var(--corner-radius);
+  font-size: 16px;
+  font-weight: 800;
+  height: 50px;
+}
+
+.ant-table-thead > tr > th {
+  font-size: 16px;
+  background-color: #323232 !important;
+  border-bottom: 1px solid #323232 !important;
```
dtfiedler, line 473: do we need `!important` ? can you explain what it's doing?

src/index.css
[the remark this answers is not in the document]
dtfiedler: do antd tables and rows support `className` props or similar?

src/index.css
```diff
@@ -440,6 +440,109 @@ input[type='text']:disabled {
...
+  background-color: inherit !important;
+}
+.ant-table-cell-ellipsis {
+  background-color: inherit;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 150px;
+}
+.ant-table-tbody > tr.ant-table-row:hover > td {
```
dtfiedler, line 542: do we need a hover effect?

src/index.css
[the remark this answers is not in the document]
dtfiedler: e.g. `rowClassName` - curious if we can set our own className and add stylings that way

src/index.css
atticusofsparta: yeah ant has some default hover styling that i need to override... it was making the rows white.

src/index.css
```diff
@@ -440,6 +440,109 @@ input[type='text']:disabled {
   height: 50px;
 }
 
+.ant-table-content > table {
+  border-spacing: 0px 10px;
+  width: 100%;
+  background-color: var(--bg-color);
+  border: none !important;
```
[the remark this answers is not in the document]
atticusofsparta: nope

src/index.css
atticusofsparta: these modifications were only taking place with the important flag

src/index.css
dtfiedler: it still seems to glitch when you hover. is there a way to fix that?

atticusofsparta: fixed

dtfiedler:
approving for now - we have some cleanup to do on our classes when it comes to v2

dtfiedler (approved):

PR #200, opened 2023-05-05
Pe 3613 update pdnt table

atticusofsparta:


dtfiedler:
remove the tooltip - and `Copied` button is very far away - put it in line of the row 
image

src/hooks/useWalletPDNTs/useWalletPDNTs.tsx
```diff
@@ -149,9 +151,26 @@ export function useWalletPDNTs(ids: ArweaveTransactionID[]) {
         className: 'white',
         ellipsis: true,
         render: (val: string) =>
-          `${val.slice(0, isMobile ? 2 : 6)}...${val.slice(
-            isMobile ? -2 : -6,
-          )}`,
+          val.length == 43 ? (
```
dtfiedler, line 154: we should be setting `targetID` to `None` or `N/A` in the table as opposed to `...` - so let's change that based on figma and do the check here when it's not equal to that value instead of just checking the length

src/hooks/useWalletPDNTs/useWalletPDNTs.tsx
```diff
@@ -149,9 +151,26 @@ export function useWalletPDNTs(ids: ArweaveTransactionID[]) {
...
+                wrapperStyle={{
+                  alignItems: 'center',
+                  margin: 'auto',
+                  fontSize: '16px',
+                }}
+              />
+            </>
+          ) : (
+            val
+          ),
```
dtfiedler, line 172:
```suggestion
          {
          if (val === "N/A") return val;
          
          return (
              <CopyTextButton
                copyText={val}
                displayText={`${val.slice(0, isMobile ? 2 : 6)}...${val.slice(
                  isMobile ? -2 : -6,
                )}`}
                size={'70%'}
                position="relative"
                wrapperStyle={{
                  alignItems: 'center',
                  margin: 'auto',
                  fontSize: '16px',
                }}
              />
          }
```

src/hooks/useWalletPDNTs/useWalletPDNTs.tsx
```diff
@@ -194,9 +212,26 @@ export function useWalletPDNTs(ids: ArweaveTransactionID[]) {
         width: '18%',
         className: 'white',
         render: (val: string) =>
-          `${val.slice(0, isMobile ? 2 : 6)}...${val.slice(
-            isMobile ? -2 : -6,
-          )}`,
+          val === 'N/A' ? (
+            val
+          ) : (
+            <>
```
dtfiedler, line 218: i don't think you need this fragment here

dtfiedler (approved):

PR #207, opened 2023-05-10
Bug Fix (PE-3522): size props had invalid size format for firefox

atticusofsparta:
image


matibat (approved):
Ahh yes, sizes must have the unit unless it is zero - LGTM

PR #212, opened 2023-05-15
Feat(PE-3653): leasing updates

atticusofsparta:


src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -0,0 +1,590 @@
...
+      }
+      if (imports) {
+        imports.map((id) =>
+          arweaveDataProvider
+            .validateTransactionTags({
+              id: id.toString(),
+              requiredTags: {
+                'Contract-Src': pdnsSourceContract.approvedANTSourceCodeTxs,
+              },
+            })
```
dtfiedler, line 123: we've decided we'll no longer be validating if an ANT contract is in an approved list - so we can likely remove this check.

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -0,0 +1,590 @@
...
+            }),
+        );
+        setValidImport(true);
+      }
+      const ids = fetchedIds.ids.concat(imports ?? []);
+
+      const contracts: [ArweaveTransactionID, PDNTContractJSON][] =
+        await Promise.all(
+          ids.map(async (id) => {
+            const state = await arweaveDataProvider.getContractState(id);
```
dtfiedler, line 138:
```suggestion
            const state = await arweaveDataProvider.getContractState<PDNTContractJSON>(id);
```

and remove the case below

src/components/inputs/text/ValidationInput/ValidationInput.tsx
```diff
@@ -118,7 +118,9 @@ function ValidationInput({
         onClick={onClick ? () => onClick() : undefined}
       >
         <div className="flex" style={{ width: '100%', position: 'relative' }}>
+          {' '}
```
dtfiedler, line 121:
```suggestion
```

src/utils/constants.ts
```diff
@@ -64,10 +64,12 @@ export const TIER_DATA: { [x: number]: string[] } = {
 };
 export const NAME_PRICE_INFO =
   'Registration fees are determined by the character length of the domain, lease duration, and what tier you choose.';
+export const PERMABUY_INFO =
```
dtfiedler, line 67: the mulitpliers here will be set in state

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
atticusofsparta: switched this to simply checking if its a smartweave contract for now. Thats about the minimum validation we should do here.

src/utils/constants.ts
atticusofsparta: updated to be a function that returns based on provided contract info, manually set for now

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
atticusofsparta: removed / switch in https://github.com/ar-io/pdns-react/pull/212/commits/dbd3df90db10dcc9610fa630ef02e2c98b92a59e

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -0,0 +1,591 @@
...
+}: {
+  walletAddress: ArweaveTransactionID;
+  selectedTokenCallback: (id: ArweaveTransactionID | undefined) => void;
+}) {
+  const arweaveDataProvider = useArweaveCompositeProvider();
+  const [{ pdnsSourceContract }] = useGlobalState();
+
+  const [searchText, setSearchText] = useState<string>();
+  const [tokens, setTokens] = useState<{
+    [x: string]: Partial<PDNTContractJSON>;
```
dtfiedler, line 35: `Partial` is too flexible, if you only want to include certain values use `Pick<PDNTContarctJSON>` or `Omit<PDNTContractjSON>`

dtfiedler: and `nameTokens` might be more clear here - ChatGPT tokens comes to mind when i see this 😆

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -0,0 +1,591 @@
...
+  }>();
+  const [loading, setLoading] = useState(false);
+  const [filteredTokens, setFilteredTokens] =
+    useState<
+      Array<{ id: string; name?: string; ticker?: string } | undefined>
+    >();
+  const [selectedToken, setSelectedToken] = useState<
+    { id: string; name: string; ticker: string } | undefined
```
dtfiedler, line 43: see comment about `Pick` vs. `Omit` and standardize `tokens` and `selectedToken` to have matching types

src/components/layout/SearchBarFooter/SearchBarFooter.tsx
```diff
@@ -12,30 +11,29 @@ function SearchBarFooter({
...
-              <div className="error-container">
-                <span className="illegal-char">{defaultText}</span>
-              </div>
-            ) : (
-              defaultText
-            )}
-          </div>
-        </>
-      ) : !isAvailable && searchResult && searchTerm ? (
+      <>
```
dtfiedler, line 14: remove
```suggestion
```

src/components/pages/Home/Home.tsx
```diff
@@ -116,6 +115,10 @@ function Home() {
                   state: `/?search=${domain}`,
                   replace: true,
                 });
+                dispatchRegisterState({
+                  type: 'reset',
+                });
+                return;
```
dtfiedler, line 121: i don't think this gets called since we navigate away before this gets called

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -0,0 +1,591 @@
...
+  >(undefined);
+  const [searching, setSearching] = useState<boolean>(false);
+  const [searchActive, setSearchActive] = useState(false);
```
karlprieb, line 46: keep the standard. Or we set every `state` type or let it infer

src/components/layout/RegisterNameForm/RegisterNameForm.tsx
```diff
@@ -7,27 +8,51 @@ import { useRegistrationState } from '../../../state/contexts/RegistrationState'
...
 
+  useEffect(() => {
+    const fees = pdnsSourceContract.fees;
+    if (domain) {
+      const newFee = calculatePDNSNamePrice({
+        domain,
+        selectedTier: tier,
+        years: leaseDuration,
+        fees,
+      });
```
dtfiedler, line 48: we also need to make sure the user has this balance before they can hit next

src/components/layout/RegisterNameForm/RegisterNameForm.tsx
```diff
@@ -113,75 +131,189 @@ function RegisterNameForm() {
...
+                    Registration Period (years)
+                  </span>
+                  <span className="text-medium faded bold center">
+                    Indefinite
+                  </span>
+                  <span
+                    className="flex flex-column white flex-center"
+                    style={{ width: '75%', fontSize: '16px', gap: '1em' }}
+                  >
+                    {permabuyInfo(fee.io, 10, 2)}
```
dtfiedler, line 251: add a comment saying these should be pulled from contract state

PR #224, opened 2023-06-02
Feat(PE-3950): update featured domains and arns cards

atticusofsparta:
updates the featured domains component and arns card (pdns card) component to match figma,

Threw in a mobile fix for the wonky footer as well.

image

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -76,6 +76,18 @@ input[type='number']:disabled {
   cursor: pointer;
   padding: 0px;
 }
+a:link {
+  text-decoration: none;
+}
+a:visited {
+  text-decoration: none;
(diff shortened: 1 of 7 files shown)


src/index.css
```diff
@@ -76,6 +76,18 @@ input[type='number']:disabled {
...
+}
+a:visited {
+  text-decoration: none;
+}
+a:hover {
+  text-decoration: none;
+}
+a:active {
+  text-decoration: none;
+}
```
dtfiedler, line 90: nit: i think you can do a:link,:visited,:hover,:active to consolidate these

dtfiedler (approved):

PR #227, opened 2023-06-06
feat(PE-3970): update searchbar design

atticusofsparta:
image

### DIFF
--- a/src/components/inputs/Search/SearchBar/SearchBar.tsx
+++ b/src/components/inputs/Search/SearchBar/SearchBar.tsx
@@ -228,10 +228,6 @@ function SearchBar(props: SearchBarProps) {
           className="accent-button center"
           onClick={_onSubmitButton}
           style={{
-            width: 130,
-            height: 50,
-            fontSize: 14,
-            padding: 0,
             marginTop: 30,
           }}
         >
--- a/src/components/inputs/Search/SearchBar/SearchBar.tsx
+++ b/src/components/inputs/Search/SearchBar/SearchBar.tsx
@@ -106,6 +106,21 @@ function SearchBar(props: SearchBarProps) {
     _onSubmit(true);
   }
 
+  const handleSearchbarBorderStyle = () => {
+    if (searchBarText) {
+      if (searchSubmitted) {
+        if (isAvailable) {
+          return { borderColor: 'var(--success-green)' };
+        } else {
+          return { borderColor: 'var(--error-red)' };
+        }
+      }
+      return { borderColor: 'white', marginBottom: 30 };
+    } else {
+      return { borderColor: '', marginBottom: 30 };
+    }
+  };
+
   return (
     <div className="searchbar-container flex-center" style={{ maxWidth: 787 }}>
       {headerElement ? (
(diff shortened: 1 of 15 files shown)


src/components/inputs/Search/SearchBar/SearchBar.tsx
```diff
@@ -123,12 +117,15 @@ function SearchBar(props: SearchBarProps) {
...
         className="searchbar"
         style={
-          isSearchValid
-            ? !searchSubmitted || showDefaultText
-              ? { borderColor: '' }
+          !searchBarText
+            ? { borderColor: '', marginBottom: 30 }
+            : isSearchValid
+            ? !searchSubmitted
+              ? { borderColor: 'white', marginBottom: 30 }
```
karlprieb, line 128: really hard to read that. Is not easy to have a `const` that will evaluate the right styles and use it here?

atticusofsparta: im confused, do you mean do the same thing but move it to a const?

dtfiedler: yeah - i think he's suggesting for such a complicated style rule, move it to a function or constant and avoid the nested ternary operators (i agree)

atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/227/commits/ef1c3d28bdea0b95c3fceabe861cf6142082e488

dtfiedler:
image

src/components/inputs/Search/SearchBar/SearchBar.tsx
```diff
@@ -123,86 +132,114 @@ function SearchBar(props: SearchBarProps) {
...
+            }}
+            onClick={() => {
+              isSearchValid && _onSubmit();
+            }}
+          >
+            <SearchIcon fill="white" width={18} height={18} />
+          </button>
         )}
       </div>
+      {searchSubmitted && isAvailable ? (
```
dtfiedler, line 226: see screenshot about when a wallet is not connected

src/components/inputs/Search/SearchBar/SearchBar.tsx
karlprieb: Great, thanks!

src/components/inputs/Search/SearchBar/SearchBar.tsx
atticusofsparta: removed in https://github.com/ar-io/pdns-react/pull/227/commits/b6d03592496f2821c7afbb3882d90237c9944b54

src/components/inputs/Search/SearchBar/SearchBar.tsx
```diff
@@ -123,86 +132,114 @@ function SearchBar(props: SearchBarProps) {
...
+        <button
+          className="accent-button center"
+          onClick={_onSubmitButton}
+          style={{
+            width: 130,
+            height: 50,
+            fontSize: 14,
+            padding: 0,
+            marginTop: 30,
```
dtfiedler, line 235: why can't we use an existing button class for this? also - noticeably moves down the feature domain on the page when this shows. not a stopper - but seems glitchy

atticusofsparta: the moving to featured domains is actually part of the figma - that state of the searchbar pushed it down while maintaining the gap.

Updated the class to be more generic with common styles.

changes in https://github.com/ar-io/pdns-react/pull/227/commits/86a8620bc0107969766621a5438b2912e2f1cf82

dtfiedler (approved):

PR #229, opened 2023-06-07
Feat(PE-3853): name token selector dropdown

atticusofsparta:


src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -1,27 +1,41 @@
...
   const arweaveDataProvider = useArweaveCompositeProvider();
+  const [{ pdnsSourceContract, walletAddress }] = useGlobalState();
 
   const [searchText, setSearchText] = useState<string>();
-  const [selectedNameToken, setSelectedNameToken] = useState<
+  const [tokens, setTokens] = useState<{
+    [x: string]: Pick<
+      PDNTContractJSON,
+      'name' | 'ticker' | 'owner' | 'controller'
+    > & { names?: string[] };
```
dtfiedler, line 31: create a type as this is used in a few spots - can be created in this component as it's not needed anywhere else

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +44,512 @@ function NameTokenSelector({
...
+        );
+      }
+      if (imports) {
+        imports.map((id) =>
+          arweaveDataProvider
+            .validateTransactionTags({
+              id: id.toString(),
+              requiredTags: {
+                'App-Name': ['SmartWeaveContract'],
+              },
```
dtfiedler, line 121: we should likely load the state and validate the schema as well

e.g. getContractState() and then `new PDNTContract(state).isValid()`

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +44,512 @@ function NameTokenSelector({
...
+        );
+        setValidImport(true);
+      }
+      const ids = fetchedIds.ids.concat(imports ?? []);
+
+      const contracts: [ArweaveTransactionID, PDNTContractJSON][] =
+        await Promise.all(
+          ids.map(async (id) => {
+            const state =
+              await arweaveDataProvider.getContractState<PDNTContractJSON>(id);
```
dtfiedler, line 138: see comment about it's worth validating the state is a valid contract using the `PDNTContract.isValid()` method

atticusofsparta: updated in https://github.com/ar-io/pdns-react/pull/229/commits/0b63d7bccbf324d95029f2a91e4a74e4e9116c0e

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
atticusofsparta: updated in https://github.com/ar-io/pdns-react/pull/229/commits/0b63d7bccbf324d95029f2a91e4a74e4e9116c0e

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
atticusofsparta: updated in https://github.com/ar-io/pdns-react/pull/229/commits/0b63d7bccbf324d95029f2a91e4a74e4e9116c0e

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +44,512 @@ function NameTokenSelector({
...
+                    .toLocaleLowerCase()
+                    .includes(query.toLocaleLowerCase())
+                : false,
+          );
+          if (queryResult.includes(true)) {
+            return { id: id, name: name ?? '', ticker: ticker ?? '' };
+          }
+        })
+        .filter((n) => n !== undefined);
+      if (!filteredResults) {
```
matibat, line 223: Same here: this is a non-nullable array. Will always return false.

matibat: Also, by applying the mapping first (which may return undefined) and then the filter, the code is inferring the type so it ends up having nullable items (even tho you filtered them out). Maybe applying the filter first, and then the mapping is the way to go.

Take a look to my chat with GPT - https://chat.openai.com/share/5ad748a1-fcab-4340-be7d-1a3edfa8b0f0

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +44,512 @@ function NameTokenSelector({
...
+                    return;
+                  }
+
+                  return (
+                    <button
+                      key={index}
+                      className="name-token-item pointer"
+                      onClick={() => {
+                        handleSetToken({
+                          id: token?.id,
```
matibat, line 478: It seems this one is non-nullable by this point
```suggestion
                          id: token.id,
```

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +44,512 @@ function NameTokenSelector({
...
-              onClick={() => setSelectedNameToken(undefined)}
             >
-              Remove
-              <CircleXIcon
-                width={'20px'}
-                height={'20px'}
-                fill={'var(--text-faded)'}
+              <Pagination
+                total={
+                  tokens && !filteredTokens
```
matibat, line 540: Same here: `tokens` is non-nullable by this point

src/components/inputs/text/NameTokenSelector/styles.css
```diff
@@ -0,0 +1,99 @@
+.custom-next-pagination {
+  display: 'flex';
+  justify-content: space-between;
+  width: 100%;
+}
+.custom-next-pagination > ul > li {
+  display: flex !important;
```
matibat, line 7: Just a comment, not requesting changes: Be careful when using `!important` because it will overwrite any other style that matches the selector and may be messy later on. The way of making a CSS rule to be applied over others is to have a more specific CSS selector.

src/components/layout/RegisterNameForm/RegisterNameForm.tsx
```diff
@@ -258,7 +258,14 @@ function RegisterNameForm() {
 
         <div className="flex flex-column" style={{ gap: '75px' }}>
           <NameTokenSelector
-            selectedTokenCallback={(id) => handlePDNTId(id?.toString())}
+            selectedTokenCallback={(id) =>
+              id
+                ? handlePDNTId(id?.toString())
```
matibat, line 263: It seems we're never passing a null value to this method. I'd suggest making the parameter of `handlePDNTId` to be non-nullable. Also, by this point id is non-nullable :)
```suggestion
                ? handlePDNTId(id.toString())
```

matibat (commented):
Some comments regarding null values 🤖

src/components/inputs/text/NameTokenSelector/styles.css
atticusofsparta: Right - i think it might be necessary here because of how ant design controls their styles using the :where selector. I tried multiple different methods, but this should *only* modify those elements belonging to the custom-next-pagination class.

src/components/layout/RegisterNameForm/RegisterNameForm.tsx
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/4de5520d092e8e98eaa9dcffe69150aac929361b

src/components/inputs/text/NameTokenSelector/styles.css
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/4de5520d092e8e98eaa9dcffe69150aac929361b

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/4de5520d092e8e98eaa9dcffe69150aac929361b

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/4de5520d092e8e98eaa9dcffe69150aac929361b

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +44,512 @@ function NameTokenSelector({
...
+            owner,
+            controller,
+            name,
+            ticker,
+            names: associatedNames,
+          },
+        };
+      }, {});
+      setTokens(newTokens);
+      if (imports && newTokens) {
```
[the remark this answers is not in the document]
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/4de5520d092e8e98eaa9dcffe69150aac929361b

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +44,512 @@ function NameTokenSelector({
...
+
+      const contracts: [ArweaveTransactionID, PDNTContractJSON][] =
+        await Promise.all(
+          ids.map(async (id) => {
+            const state =
+              await arweaveDataProvider.getContractState<PDNTContractJSON>(id);
+            return [id, state];
+          }),
+        );
+      if (!contracts) {
```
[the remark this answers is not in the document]
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/4de5520d092e8e98eaa9dcffe69150aac929361b

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +44,512 @@ function NameTokenSelector({
...
+  ) {
+    try {
+      setLoading(true);
+      if (!address) {
+        throw new Error('No address provided');
+      }
+      const fetchedIds = await arweaveDataProvider.getContractsForWallet(
+        address,
+      );
+      if (!fetchedIds.ids) {
```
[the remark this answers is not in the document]
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/4de5520d092e8e98eaa9dcffe69150aac929361b

src/components/inputs/text/NameTokenSelector/styles.css
matibat: Nothing to change here 👍

matibat (dismissed):
Looks good to me now - thanks!

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +47,513 @@ function NameTokenSelector({
...
+  }) {
+    setSearchText('');
+    setFilteredTokens(undefined);
+
+    if (id === undefined) {
+      if (selectedToken === undefined) {
+        setSearchActive(false);
+      }
+      return;
+    }
```
dtfiedler, line 250: i'm a bit confused by this logic - if `id` is `undefined` but `selectedToken` is not we return? can we consolidate to something like:
```suggestion
if (!id && !selectedToken) {
   setSearchActive(false);
   return;
}
```

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,177 +47,513 @@ function NameTokenSelector({
...
-    <div
-      className="flex flex-column"
-      style={{
-        position: 'relative',
-        height: 'fit-content',
-        maxHeight: '400px',
-      }}
-    >
-      {/* input wrapper */}
+    <>
```
dtfiedler, line 289: extraneous js fragment

src/hooks/useArweaveCompositeProvider/useArweaveCompositeProvider.tsx
```diff
@@ -10,7 +10,7 @@ import eventEmitter from '../../utils/events';
 
 const PDNS_SERVICE_API =
   process.env.VITE_ARNS_SERVICE_API ?? 'https://dev.pdns.app';
-const ARWEAVE_HOST = process.env.VITE_ARWEAVE_HOST ?? 'ar-io.dev';
+const ARWEAVE_HOST = process.env.VITE_ARWEAVE_HOST ?? 'arweave.net';
```
dtfiedler, line 13: is this intentional?

atticusofsparta: oops, no

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
atticusofsparta: it is required to verify id is there before proceeding which needs to be an independent check.

I moved this to just the id check, wrapped in try-catch-finally

src/hooks/useArweaveCompositeProvider/useArweaveCompositeProvider.tsx
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/770ee8aa94514c233cac10ecee55f3a26abc37a7

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
atticusofsparta: fixed in https://github.com/ar-io/pdns-react/pull/229/commits/770ee8aa94514c233cac10ecee55f3a26abc37a7

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,85 +47,281 @@ function NameTokenSelector({
...
+        );
+      }
+      if (imports?.length) {
+        imports.map(async (id) => {
+          await arweaveDataProvider.validateTransactionTags({
+            id: id.toString(),
+            requiredTags: {
+              'App-Name': ['SmartWeaveContract'],
+            },
+          });
```
dtfiedler, line 117: we're not doing anything with the response here - why not?

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,85 +47,281 @@ function NameTokenSelector({
...
+          }
+          return [id, state];
+        }),
+      );
+      if (!contracts.length) {
+        throw new Error('Unable to get details for Name Tokens');
+      }
+      const newTokens: NameTokenDetails = contracts.reduce(
+        (result, contract) => {
+          if (!contract) {
```
dtfiedler, line 151: why would contract be potentially undefined?

dtfiedler: ah,  i see - from the potential null above.

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,85 +47,281 @@ function NameTokenSelector({
...
+      > = await Promise.all(
+        ids.map(async (id) => {
+          const state =
+            await arweaveDataProvider.getContractState<PDNTContractJSON>(id);
+          if (!state) {
+            return;
+          }
+          if (!new PDNTContract(state).isValid()) {
+            return;
+          }
```
dtfiedler, line 142: nit: can combine these or throw error

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,85 +47,281 @@ function NameTokenSelector({
...
+              term &&
+              term.toString().toLowerCase().includes(query.toLowerCase()),
+          );
+          return queryResult;
+        })
+        .map((id) => {
+          const { name, ticker } = tokens[id];
+          return { id, name: name ?? '', ticker: ticker ?? '' };
+        })
+        .filter((n) => n !== undefined);
```
dtfiedler, line 218:
```suggestion
        .filter((n) => !!n); // handles all falsey values instead of just undefined
```

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -30,85 +47,281 @@ function NameTokenSelector({
...
+      if (!filteredResults.length) {
+        throw new Error('No ANT tokens found for that search');
```
dtfiedler, line 220: do we actually want to throw an error here?

atticusofsparta: For the moment yes.

dtfiedler (dismissed):

src/components/inputs/text/NameTokenSelector/NameTokenSelector.tsx
```diff
@@ -117,7 +330,7 @@ function NameTokenSelector({
                 arweaveDataProvider.validateTransactionTags({
                   id,
                   requiredTags: {
-                    'App-Name': ['SmartWeaveContract'],
+                    'Contract-Src': pdnsSourceContract.approvedANTSourceCodeTxs,
```
dtfiedler, line 333: no longer validating src codes

PR #240, opened 2023-06-15
FIX(PE-4027): bug version is showing

atticusofsparta:
Removed version when not running local


dtfiedler:
the intent of the version is so we know what is deployed on `arns.app` - we can put this behind a debug flag, but regardless we'll want it once deployed so we can validate what versions people are seeing issues on

PR #493, opened 2024-07-15
fix(publish): optimize permaweb deploy

atticusofsparta:


package.json
```diff
@@ -5,7 +5,7 @@
   "homepage": ".",
   "scripts": {
     "build": "yarn clean && tsc --build tsconfig.json && NODE_OPTIONS=--max-old-space-size=32768 vite build",
-    "publish:arweave": "yarn build && permaweb-deploy --ant-process $DEPLOY_ANT_PROCESS_ID ",
+    "publish:arweave": "permaweb-deploy --ant-process $DEPLOY_ANT_PROCESS_ID ",
```
kunstmusik, line 8: This should be reverted, otherwise it won't build before deploying in CI.

atticusofsparta: the action is configured to build before calling this

kunstmusik: It's dependence on the build action doesn't mean this action has access to the data from the previous job. It would need to pass artifacts between jobs (see https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts).

kunstmusik: Also, we should force a clean/build as part of deploy for package.json so that anyone doing a deploy (i.e., locally) will always have a clean build.

kunstmusik (dismissed):

kunstmusik (approved):

PR #502, opened 2024-07-23
fix(PE-6486): Capture errors from arns emitter

atticusofsparta:


src/hooks/useWalletANTs/useWalletANTs.tsx
```diff
@@ -440,6 +442,7 @@ export function useWalletANTs() {
         dispatch: dispatchArNSState,
         emitter: arnsEmitter,
         walletAddress,
+        registryId: arnsContractId,
```
kunstmusik, line 445: I think we've got three different names for the same thing now:

* registryId
* arnsContractId
* ioProcessId

If these are all the same thing, we should settle one a single name.

src/state/actions/dispatchArNSUpdate.ts
```diff
@@ -40,6 +44,37 @@ export function dispatchArNSUpdate({
       payload: totalIds,
     });
   });
+  const errorHandler = (e: string) => {
+    if (e.startsWith('Error getting ArNS records')) {
+      eventEmitter.emit(
+        'error',
+        new Error('Unable to load ArNS records. Please refresh to try again.'),
+      );
+      // Behaviour note: this will only stop the loading UI, it will still display the
```
kunstmusik, line 53: This comment just ends with "the", seems like some more is meant to be mentioned?

src/state/actions/dispatchArNSUpdate.ts
```diff
@@ -40,6 +44,37 @@ export function dispatchArNSUpdate({
...
+        tags: {
+          walletAddress: walletAddress.toString(),
+          registryId: registryId.toString(),
+        },
+      });
+    } else if (!e.includes('does not support provided action.')) {
+      eventEmitter.emit('error', new Error(e));
+    }
+  };
+  // error listnener handles timeout
```
kunstmusik, line 74: typo: listener

src/state/actions/dispatchArNSUpdate.ts
```diff
@@ -51,5 +86,9 @@ export function dispatchArNSUpdate({
     });
   });
 
-  emitter.fetchProcessesOwnedByWallet({ address: walletAddress.toString() });
+  emitter
+    .fetchProcessesOwnedByWallet({ address: walletAddress.toString() })
+    .catch((e) =>
+      errorHandler('Error getting processes owned by wallet: ' + e.message),
```
kunstmusik, line 92: Would it be clearer to the user to mention "ANT processes"?

src/state/actions/dispatchArNSUpdate.ts
atticusofsparta: Realized the comment was incorrect anyways, removing.

src/state/actions/dispatchArNSUpdate.ts
atticusofsparta: Specified assets instead sinces both domains and ants are affected

src/state/actions/dispatchArIOContract.ts
```diff
@@ -1,17 +1,24 @@
...
   contract: AoIORead | AoIOWrite;
+  ioProcessId: ArweaveTransactionID;
   dispatch: Dispatch<GlobalAction>;
 }) {
   dispatch({
     type: 'setArIOContract',
     payload: contract,
   });
+  dispatch({
+    type: 'setioProcessId',
```
kunstmusik, line 21: looks strange, suggest to change this to "setIoProcessId" and in reducer

atticusofsparta: fixed

kunstmusik (approved):

PR #521, opened 2024-08-09
feat(PE-6553): Evolve ANTs workflow

atticusofsparta:


package.json
```diff
@@ -21,12 +21,11 @@
   },
   "dependencies": {
     "@ant-design/icons": "5.3.2",
-    "@ar.io/sdk": "2.1.0",
+    "@ar.io/sdk": "2.1.0-alpha.11",
```
kunstmusik, line 24: Looks like this went backwards in version?

package.json
```diff
@@ -65,6 +65,8 @@
     "@typescript-eslint/eslint-plugin": "^5.42.1",
     "@typescript-eslint/parser": "^6.19.0",
     "@vitejs/plugin-react": "^4.0.0",
+    "arconnect": "^1.0.3",
```
kunstmusik, line 68: Shouldn't this be a regular dependency vs. dev dependency?

.yarnrc.yml
dtfiedler, line 1: intent here? was this intentional? we've used yarn berry before but moved away from it as it's overhead wasn't giving us much

dtfiedler (changes requested):
This PR is out of date and has some un-adressed comments

package.json
atticusofsparta: the name is misleading, its a types package

.yarnrc.yml
atticusofsparta: I was running into issues with the tailwind install I think relating to yarn, ill revert and see what the issue was here again

atticusofsparta: It was a build issue with ant-design dep issues that I was able to fix without bumping yarn, so reverted to 1.22

src/components/inputs/buttons/CopyTextButton/CopyTextButton.tsx
```diff
@@ -71,18 +71,21 @@ function CopyTextButton({
             padding: '4px 8px',
           }}
         >
-          <CopyIcon
-            className="flex"
-            height={size ?? '15px'}
-            width={size ?? '15px'}
-            fill={'inherit'}
+          <div
+            className="flex tems-center justify-center"
```
kunstmusik, line 75: typo: items-center

package.json
```diff
@@ -107,5 +115,6 @@
     "**/*.{ts,tsx,js,md,json}": [
       "prettier --write ."
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22"
```
dtfiedler, line 119: any particular reason for introducing this? if so can we separate it into a specific PR. Jamming these things in with feature changes makes creates noise

src/components/layout/ExtendLease/ExtendLease.tsx
```diff
@@ -159,7 +159,7 @@ function ExtendLease() {
           className="flex flex-row center"
           style={{ justifyContent: 'space-between' }}
         >
-          <h1 className="white">Extend Lease</h1>
+          <h1 className="white text-2xl">Extend Lease</h1>
```
kunstmusik, line 162: Not sure it's intentional, but this is smaller than what was previously in prod. (Renders as 24px size here). It's odd as 2xl should be 32px with a 16px base font size; unsure if the h1 is interfering somehow.

kunstmusik: Ah, it is 24px for 2xl. To get 32px use text-[2rem]

src/components/layout/UpgradeUndernames/UpgradeUndernames.tsx
```diff
@@ -113,7 +113,7 @@ function UpgradeUndernames() {
   return (
     <div className="page center">
       <div className="flex flex-column" style={{ maxWidth: '1000px' }}>
-        <h1 className="flex white">Increase Undernames</h1>
+        <h1 className="flex white text-2xl">Increase Undernames</h1>
```
kunstmusik, line 116: Modify to text-[2rem]

package.json
atticusofsparta: no - will remove. After effect from the yarn berry thing.

dtfiedler:
The UI looks good, however I had some issues.

When I first logged in `Upgrade Ants` said I would update 16 ANTS. I clicked `Approve` and `Updated`. The loading indicator showed the progress, and showed the error for one ANT that failed. Fair enough. When I returned to `Manage` Page, I clicked the`Upgrade ANTs` again expecting it to show `1 ANT needs to be updated` instead it still showed `16 ANTs need to be updated`. I went ahead anyway and saw the requests for 16 ANTs go through again, not just the 1.

image

So questions:
- how are you tracking which needs to be upgraded? Can we show that on  a column in the table (bases on that criteria)
- how can i individually update one ANT? - is that just another button on the row that says upgrade?

kunstmusik (approved):

PR #532, opened 2024-09-04
feat(PE-6612): update manage asset table

atticusofsparta:


.eslintrc
```diff
@@ -36,6 +36,7 @@
     "@typescript-eslint/no-unused-vars": ["error", {}],
     "@typescript-eslint/no-explicit-any": "off",
     "no-useless-escape": "off",
-    "react/no-children-prop": "off"
+    "react/no-children-prop": "off",
+    "react/prop-types": "off"
```
kunstmusik, line 40: Why is react/prop-types off here? Seems like it'll hide problems that should be fixed.

src/utils/searchUtils/searchUtils.ts
```diff
@@ -157,3 +157,23 @@ export function doAntsRequireUpdate({
 
   return getAntsRequiringUpdate({ ants, luaSourceTx }).length > 0;
 }
+
+export function camelToReadable(camel: string) {
```
kunstmusik, line 161: Would be good to have a unit test for this

src/utils/searchUtils/searchUtils.ts
```diff
@@ -157,3 +157,23 @@ export function doAntsRequireUpdate({
...
+  const words = camel.replace(/([A-Z])/g, ' $1').toLowerCase();
+  return words.charAt(0).toUpperCase() + words.slice(1);
+}
+
+export function getOwnershipStatus(
```
kunstmusik, line 166: prefer a single expression of

```
return (owner === walletAddress) ? 
         'owner' : 
         (walletAddress && controllers.includes(walletAddress)) ? 
         'controller' : 
         undefined;
```

src/utils/common/common.tsx
```diff
@@ -10,6 +10,23 @@ export function formatDate(epochMs: number): string {
   return new Date(epochMs).toISOString().split('T')[0];
 }
 
+export function formatVerboseDate(timestamp: number | string): string {
```
kunstmusik, line 13: would be good to have a comment here with an example of the expected date string output; also a unit test would be good

src/state/contexts/ModalState.tsx
```diff
@@ -5,6 +5,7 @@ import { ModalAction } from '../reducers/ModalReducer';
 
 export type ModalState = {
   showUpgradeAntModal: boolean;
+  showEditUndernameModal: boolean;
```
kunstmusik, line 8: Not for this PR, but we should discuss the way modal's are done in Network Portal and see about a similar pattern for this app in the future.

src/components/pages/Manage/Manage.tsx
```diff
@@ -1,55 +1,32 @@
...
-  }, [path]);
-
-  function updatePage(page: number) {
-    setTablePage(page);
-  }
+  const [search, setSearch] = useState<string>('');
 
   return (
-    <div className="page" style={{ overflow: 'auto' }}>
+    <div className="page" style={{ overflow: 'auto', paddingTop: '0px' }}>
```
kunstmusik, line 29: maybe go ahead and switch this to tailwind and use `className="page pt-0 overflow-auto"`

src/components/layout/Layout/Layout.tsx
```diff
@@ -38,11 +41,25 @@ function Layout() {
       >
         <NavBar />
       </div>
+
+      <Progress
```
kunstmusik, line 45: This has a strange effect when load or reload completes, the bar animates from full all the way back down to 0. It's a bit jarring, IMO, and the bar should just disappear.

src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -0,0 +1,467 @@
...
+        };
+        newTableData.push(data);
+      });
+
+      setTableData(newTableData);
+    }
+  }, [domainData, loading]);
+
+  useEffect(() => {
+    if (filter) {
```
kunstmusik, line 145: I think there may be something off here. If I put in a text that triggers the filter, then hit backspace repeatedly to clear it out, the filter updates and clears out. If instead I select the text and delete it or select and cut the text with cmd-x, the table does not update.

.eslintrc
atticusofsparta: It was requiring prop types in implementations as well as definitions. Typescript already covers this.

kunstmusik (approved):

PR #534, opened 2024-09-10
feat(PE-6682): rebuilt search

atticusofsparta:


tests/search/HomeSearch.test.tsx
```diff
@@ -0,0 +1,123 @@
+import HomeSearch from '@src/components/inputs/Search/HomeSearch';
+import { decodeDomainToASCII, lowerCaseDomain } from '@src/utils';
+import { cleanup, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { act } from 'react';
+
+jest.mock('@src/hooks/useArNSRegistryDomains', () => ({
+  useArNSRegistryDomains: jest.fn().mockReturnValue({
```
atticusofsparta, line 8: @kunstmusik I think here is where its failing - its not providing the data to the tests

src/components/inputs/Search/SearchBar/SearchBar.test.tsx
dtfiedler, line 1: 👋 

looking back - i recommended a bad approach for this component. i would've not had include the header and footer components that caused some churn and refactoring as we evolved the search functionality. i also think we should have used tanstack from the start. thanks to @kunstmusik wisdom and experience, hopefully we turn the page on what was a clunky and iterative implementation (again, due to my poor direction/recs after being out of touch with react for a few years) to something more scaleable and maintainable

src/components/pages/Home/Home.tsx
```diff
@@ -96,17 +35,11 @@ function Home() {
...
-          reservedFor: reservedFor,
-          antId: antID,
-          domainName: lowerCaseDomain(domain),
-        }) && !isValidatingRegistration ? (
-          <FeaturedDomains />
-        ) : (
-          <></>
-        )}
+        <div className="flex flex-col w-full max-w-[800px] h-[200px]">
+          <HomeSearch />
```
dtfiedler, line 39: 👍

src/hooks/useArNSDomainPriceList.tsx
```diff
@@ -0,0 +1,53 @@
...
+              purchaseType: 'lease',
+            }),
+            arioContract.getTokenCost({
+              ...sharedOptions,
+              purchaseType: 'permabuy',
+            }),
+          ]);
+          prices.lease = leasePrice;
+          prices.buy = buyPrice;
+        }
```
dtfiedler, line 42: lt's get something in the SDK and contract that supports this vs slamming in here

src/hooks/useArNSDomainPriceList.tsx
```diff
@@ -0,0 +1,53 @@
...
+      } catch (error) {
+        eventEmitter.emit('error', {
+          message: `Failed to fetch price for domain: ${domain}`,
+          name: 'Domain Price',
+        });
+      }
+      return prices;
+    },
+    staleTime: 60 * 1000, // 1 minute
```
dtfiedler, line 51: you can trust prices for up to a full demand factor period (1-day) - so i'd say it's okay to cache this to something more like 5-15 mins

src/utils/common/common.tsx
```diff
@@ -9,6 +10,28 @@ import { fromB64Url } from '../encodings';
...
+    Date.now(),
+    endTimestamp,
+  );
+  return (
+    <>
+      {' '}
+      {leaseDuration} year
+      {leaseDuration > 1 ? 's' : ''} &nbsp;
+      <span style={{ color: 'var(--text-grey)' }}>
+        (est. expiry {+endTimestamp ? formatDate(endTimestamp) : 'N/A'})
```
dtfiedler, line 30: small nit: est. expiration is more clear

src/utils/network.ts
```diff
@@ -44,14 +44,15 @@ export const queryClient = new QueryClient({
...
 
-export function buildAntStateQuery({ processId }: { processId: string }): {
+export function buildAntStateQuery({ processId }: { processId?: string }): {
   queryKey: ['ant', string];
   queryFn: () => Promise<AoANTState | null>;
   staleTime: number;
 } {
   return {
-    queryKey: ['ant', processId],
+    queryKey: ['ant', processId ?? 'null'],
```
dtfiedler, line 53: why save - `null` as a string instead of just ''

src/utils/searchUtils/searchUtils.ts
```diff
@@ -124,6 +124,9 @@ export function sleep(ms: number) {
 }
 
 export function lowerCaseDomain(domain: string) {
+  if (domain.includes('%')) {
+    domain = decodeURIComponent(domain);
+  }
```
dtfiedler, line 129: can you not just always decode this?

tests/search/HomeSearch.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  test('renders correctly', () => {
+    expect(renderSearchBar()).toMatchSnapshot();
+  });
+
+  test('handles a capitalized name correctly', async () => {
+    const domain = 'ARDRIVE';
+
+    await userEvent.type(searchInput, domain);
+    console.log(searchInput.value);
```
dtfiedler, line 62:
```suggestion
```

tests/search/HomeSearch.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+          purchasePrice: 0,
+          name: 'xn--go8h6v',
+        },
+      ],
+    },
+    isLoading: false,
+  }),
+}));
+
+describe('HomeSearch', () => {
```
dtfiedler, line 35: before we invest more time on these formatted test, if they are not on the ideal testing pattern path we want to get to (e.g. playwright) then let's focus on migrating to those tools and adding tests after.

dtfiedler (dismissed):
several nits - but the UX is solid and didn't have any obvious issues. excited to hear feedback on this.

src/utils/common/common.tsx
```diff
@@ -9,6 +10,28 @@ import { fromB64Url } from '../encodings';
 export function formatDate(epochMs: number): string {
   return new Date(epochMs).toISOString().split('T')[0];
 }
+
+export function formatLeaseDurationFromEndTimestamp(
```
kunstmusik, line 14: Odd to see this as a utility function when it's essentially a React.FC. I'd rename it with a capitalized name (e.g., LeaseDuration) and move it into its own component file.

src/components/inputs/Search/HomeSearch.tsx
```diff
@@ -0,0 +1,290 @@
...
+import { ChevronRight, XIcon } from 'lucide-react';
+import { useEffect, useState } from 'react';
+import { useNavigate } from 'react-router-dom';
+
+import arioLoading from '../../icons/ario-spinner.json';
+import DomainSearch from '../DomainSearch';
+import DomainPiceList from './DomainPriceList';
+
+const defaultValidations = {
+  ['Max. 51 characters']: false,
```
kunstmusik, line 17: extract these keys as constants since they're repeated below

src/components/inputs/Search/HomeSearch.tsx
```diff
@@ -0,0 +1,290 @@
...
+            validDomain = false;
+            const specialChars =
+              decodeDomainToASCII(safeDomain).match(/[^a-zA-Z0-9-]/g);
+            setValidationError(`remove '${specialChars?.join("', '")}'`);
+          }
+          return;
+        case 'Dashes cannot be leading or trailing':
+          if (!isValid && validDomain) {
+            validDomain = false;
+            setValidationError('dashes cannot be leading or trailing');
```
kunstmusik, line 68: odd to have "Dashes" and then here "dashes".

kunstmusik:
General note: should add a debounce to key input so the search isn't too jarring

src/components/inputs/Search/SearchBar/SearchBar.test.tsx
atticusofsparta: Yeah I really like the new one. The colocation and new data provider hooks make it really easy to test.

Also with the DomainSearch internally implemented as an unstyled component we can reimplement in the nav search later with composed search functionality for user assets.

src/hooks/useArNSDomainPriceList.tsx
atticusofsparta: ahhh good to know

atticusofsparta: Will note to include when the next DF update is in the pricelist ticket

src/utils/network.ts
atticusofsparta: Actually, I should update this to require process id I think

tests/search/HomeSearch.test.tsx
atticusofsparta: What we have here is a good format for unit tests I think that we could use in conjunction with playwright, but honestly I have used playwright much so it may indeed be redundant.

thiagocarvalhodev (dismissed):

thiagocarvalhodev (approved):

PR #535, opened 2024-09-11
fix(PE-6492): add network configurations and update old configs

atticusofsparta:


src/components/modals/ant-management/UpgradeAntModal/UpgradeAntModal.tsx
```diff
@@ -98,6 +100,11 @@ function UpgradeAntModal({
           processId: antId,
           luaCodeTxId: ANT_LUA_ID,
           signer,
+          ao: connect({
```
kunstmusik, line 103: Is it a problem to cache ao outside of this loop and share or cache globally and share?

Also seeing CU_URL, MU_URL, and SCHEDULER makes it look like global constants. Would rename those to something like cuUrl, muUrl, and scheduler.

src/components/pages/Transaction/TransactionReview.tsx
```diff
@@ -111,6 +113,12 @@ function TransactionReview() {
         processId: ioProcessId,
         dispatch: dispatchTransactionState,
         signer: wallet?.arconnectSigner,
+        ao: connect({
+          GATEWAY_URL: 'https://' + gateway,
```
kunstmusik, line 117: same comment here about caching ao

src/components/modals/ant-management/UpgradeAntModal/UpgradeAntModal.tsx
kunstmusik: Nevermind about the casing for CU_URL, etc., I see that's a part of aoconnect. :(

kunstmusik: But still curious about caching ao :)

src/utils/constants.ts
```diff
@@ -69,12 +72,30 @@ export const EMAIL_REGEX = new RegExp(
 );
 export const TTL_SECONDS_REGEX = new RegExp('^[0-9]{3,7}$');
 export const TTL_SECONDS_ENTRY_REGEX = new RegExp('^[0-9]{1,7}$');
+
+export const FQDN_REGEX = new RegExp(
+  '^(?:(?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{1,63}$',
+);
 export const ARNS_REGISTRY_ADDRESS = new ArweaveTransactionID(
```
kunstmusik, line 79: odd here that ARNS_REGISTRY_ADDRESS defaults to testnet, whiel IO_PROCESS_ID defaults to devnet

src/components/modals/ant-management/UpgradeAntModal/UpgradeAntModal.tsx
atticusofsparta: They are global constants.

src/utils/constants.ts
atticusofsparta: ARNS_REGISTRY_ADDRESS was a smartweave var, removed it

src/components/modals/ant-management/UpgradeAntModal/UpgradeAntModal.tsx
kunstmusik: They're not, they're key names to the object passed into connect. We can't change that. i'm more concerned about caching ao vs. having to recreate each time.

kunstmusik (approved):

PR #542, opened 2024-09-17
Featured domains cards update demo

atticusofsparta:


dtfiedler:
Is this just exploratory? If so, can we mark as Draft?

atticusofsparta:
@dtfiedler yeah this was just me messing around looking for something that might be nicer for the front page. Ill close it.

PR #547, opened 2024-09-23
fix(PE-6763): fixed refresh on tab switch

atticusofsparta:


src/state/actions/dispatchArIOInteraction.ts
```diff
@@ -57,6 +60,22 @@ export default async function dispatchArIOInteraction({
             ao: ao,
             scheduler: scheduler,
           });
+          const antRegistry = ANTRegistry.init({
+            signer,
+            processId: ANT_REGISTRY_ID,
+          }) as AoANTRegistryWrite;
```
dtfiedler, line 66: why do we need the cast if signer is provided?

src/state/actions/dispatchArIOInteraction.ts
```diff
@@ -57,6 +60,22 @@ export default async function dispatchArIOInteraction({
...
+          await antRegistry
+            .register({
+              processId: antProcessId,
+            })
+            .catch((error) => {
+              eventEmitter.emit(
+                'error',
+                new Error(
+                  `Failed to register ANT process: ${error}. You may need to manually register the process`,
```
kunstmusik, line 75: Do we have a way for non-dev users to manually register?

kunstmusik: Note: this needs to get sorted out in a future PR as we're telling users they have a path to do something but then not really having one.

kunstmusik (approved):

PR #548, opened 2024-09-24
fix(PE-6820): added devtool for viewing as other user and also added fix for decode tags in hook

atticusofsparta:


src/utils/searchUtils/searchUtils.ts
```diff
@@ -144,7 +144,11 @@ export function getAntsRequiringUpdate({
     luaSourceTx.id,
     luaSourceTx?.tags?.find((tag) => tag.name == 'Original-Tx-Id')?.value,
   ];
-
+  console.log(
+    'luaSourceTx',
+    luaSourceTx.tags.map((tag) => ({ [tag.name]: tag.value })),
+  );
+  console.log('acceptableIds', acceptableIds);
```
dtfiedler, line 151:
```suggestion
```

kunstmusik (approved):

PR #555, opened 2024-10-01
fix(PE-6831): update individual ant in state when performing an interaction

atticusofsparta:


src/components/modals/ant-management/UpgradeAntModal/UpgradeAntModal.tsx
```diff
@@ -262,16 +231,16 @@ function UpgradeAntModal({
...
-                    progress < 0
+                    !upgrading
                       ? 'bg-primary-thin text-primary'
                       : 'bg-link text-white hover:bg-primary hover:text-black'
                   } `
             } w-full rounded-b-lg p-3 transition-all`}
             onClick={() => upgradeAnts()}
           >
-            {!accepted && progress < 0
+            {!accepted && !upgrading
```
kunstmusik, line 241:
```suggestion
            {!accepted
```

kunstmusik: !upgrading doesn't affect this calculation since you can't ever get into a situation of !accepted and upgrading. Remove to make this less complicated.

kunstmusik (approved):

PR #556, opened 2024-10-02
fix(hotfix for lua id): bump sdk with new lua id and update constant

atticusofsparta:


src/utils/constants.ts
```diff
@@ -84,7 +88,7 @@ export const URL_REGEX = new RegExp(
 export const IO_PROCESS_ID =
   process.env.VITE_IO_PROCESS_ID || ioDevnetProcessId;
 
-export const DEFAULT_ANT_LUA_ID = 'RuoUVJOCJOvSfvvi_tn0UPirQxlYdC4_odqmORASP8g'; // ANT_LUA_ID;
+export const DEFAULT_ANT_LUA_ID = ANT_LUA_ID; // ANT_LUA_ID;
```
dtfiedler, line 91: i want 100% confirmation this is the correct source code. hard code it if necessary

`pOh2yupSaQCrLI_-ah8tVTiusUdVNTxxeWTQQHNdf30`

PR #561, opened 2024-10-03
fix(PE-6788): update docs link in landing page and publish new version

atticusofsparta:


dtfiedler (approved):

dtfiedler:
why do we need this here if ANTs set it by default?

PR #568, opened 2024-10-04
feat(PE-6726): settings page

atticusofsparta:


src/components/inputs/toggle/Toggle.tsx
```diff
@@ -0,0 +1,74 @@
...
+  className,
+  inputClassName,
+}: {
+  leftLabel: string;
+  rightLabel: string;
+  check: (b: boolean) => void;
+  className?: string;
+  inputClassName?: string;
+}) {
+  const [checked, setChecked] = useState(false);
```
kunstmusik, line 17: Rather than have a local state and a setter coming in, think it'd be better for this component to take in checked and setChecked and remove useState here. It'd make it more inline with other input components.

src/components/inputs/toggle/Toggle.tsx
```diff
@@ -0,0 +1,74 @@
...
+          }}
+        >
+          <span
+            className={`z-10 pl-2 transition-all ${
+              !checked ? 'text-black' : 'text-primary'
+            }`}
+          >
+            {leftLabel}
+          </span>
+          <motion.div
```
kunstmusik, line 51: Nice, been meaning to look at framer-motion.

src/components/pages/Settings/ArNSSettings.tsx
```diff
@@ -0,0 +1,171 @@
...
+      });
+
+      dispatchGlobalState({
+        type: 'setGateway',
+        payload: { gateway, provider },
+      });
+    }
+  }
+
+  const labelClass =
```
kunstmusik, line 74: I generally find it easier to see the layout when the tailwind classes are inline with the elemtns.

src/components/pages/Settings/SettingsLayout.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <div className="flex flex-col w-full h-full px-[100px] py-[30px]">
+        <div className="flex flex-col w-full rounded-t-xl bg-metallic-grey border-x-[1px] border-t-[1px] border-dark-grey p-2">
+          <h1
+            className="flex flex-row text-white text-xl pl-3 "
+            style={{ gap: '15px' }}
+          >
+            <Link
+              to={'/settings'}
+              className={
+                location?.pathname?.split('/')?.at(-1) == 'settings'
```
kunstmusik, line 39: I'd refactor this location check to a variable as it's being repeated multiple times below. 

Sidenote: I haven't been using it, but Tailwind supports data-attributes which might work here: https://tailwindcss.com/docs/hover-focus-and-other-states#data-attributes

src/components/inputs/toggle/Toggle.tsx
atticusofsparta: recommend checking this one out, its a more serious component I put together as a PoC for the domain cards. Might use it on a grid view for manage assets. https://github.com/ar-io/arns-react/pull/542

src/components/inputs/toggle/Toggle.tsx
atticusofsparta: Decided to remove this component completely for now per design recs

src/components/pages/Settings/ArNSSettings.tsx
atticusofsparta: You can see it in the code when you hover over the variable - not as friendly for reviewing but its overall better DX for me.

src/components/pages/Settings/SettingsLayout.tsx
atticusofsparta: I refactored - but did not use the data attribute. In this particular case I think the JSX approach is cleaner.

If I had a non binary value that would require a multiline JSX case then I would probably use the data attribute though. Pretty cool that tailwind allows that.

src/App.tsx
```diff
@@ -52,6 +52,20 @@ const UpgradeUndernames = React.lazy(
   () => import('./components/layout/UpgradeUndernames/UpgradeUndernames'),
 );
 
+const SettingsLayout = React.lazy(
```
kunstmusik, line 55: One thing that's strange is when you go to settings, then click between the panels, first time it's lazy loading and there's a visual glitch for loading that's a bit jarring. Would be nice to work this out in a way to avoid that. (Not a blocker for release.)

kunstmusik (approved):

kunstmusik (approved):

PR #578, opened 2024-10-23
fix(PE-6915/PE-7015): upgrade undernames fixes

atticusofsparta:


dtfiedler (approved):

dtfiedler (approved):

src/components/layout/UpgradeUndernames/UpgradeUndernames.tsx
```diff
@@ -174,7 +148,7 @@ function UpgradeUndernames() {
           }}
           ioRequired={true}
           fee={{
-            [ioTicker]: fee,
+            [ioTicker]: fee ? new mIOToken(fee).toIO().valueOf() : undefined,
```
kunstmusik, line 151: There seems to be a bug which might be caused by the above checking falsey rather than explicitly for undefined. If you set the number of undernames to increase to 0 it gets stuck on "Calculating tIO". This happens if you load the page and it's 0, or if you edit and go back to 0.

kunstmusik (approved):

PR #581, opened 2024-10-23
fix(PE-7011): fix target id on spawn

atticusofsparta:


src/utils/transactionUtils/transactionUtils.tsx
```diff
@@ -694,3 +695,34 @@ export function userHasSufficientBalance<T extends Record<string, number>>({
...
+        ttlSeconds: 3600,
+      },
+    },
+    ...state,
+  };
+}
+
+export function createAntStateForOwner(owner: string, targetId: string) {
+  return createDefaultAntState({
+    owner: owner.toString(),
```
kunstmusik, line 718: toString() not necessary here in this function when it's already a string argument

src/state/actions/dispatchArIOInteraction.ts
```diff
@@ -58,8 +58,16 @@ export default async function dispatchArIOInteraction({
...
+            payload.state ||
+            (payload.targetId
+              ? createAntStateForOwner(
+                  owner.toString(),
+                  payload?.targetId?.toString(),
+                )
+              : undefined);
           antProcessId = await spawnANT({
-            state: payload.state,
+            state,
```
kunstmusik, line 70: Do we ever want to spawn an ant where state may be undefined?

src/state/actions/dispatchArIOInteraction.ts
```diff
@@ -58,8 +58,16 @@ export default async function dispatchArIOInteraction({
         let antProcessId: string = payload.processId;
 
         if (antProcessId === 'atomic') {
+          const state =
+            payload.state ||
+            (payload.targetId
```
kunstmusik, line 63: Seems odd here to check payload.targetId then later use payload?.targetId?. Can this be condensed down to:

```
 const state =
            payload.state ||
            createAntStateForOwner(
                  owner.toString(),
                  payload?.targetId?.toString(),
                );
                ```

File: src/state/actions/dispatchArIOInteraction.ts (hunk 2 above)
Comment (atticusofsparta) [reply src/state/actions/dispatchArIOInteraction.ts:70]: Yup

File: src/utils/transactionUtils/transactionUtils.tsx [hunk 4]
```diff
@@ -694,3 +695,34 @@ export function userHasSufficientBalance<T extends Record<string, number>>({
...
+        ttlSeconds: 3600,
+      },
+    },
+    ...state,
+  };
+}
+
+export function createAntStateForOwner(owner: string, targetId?: string) {
+  return createDefaultAntState({
+    owner: owner.toString(),
```
kunstmusik, line 718: still owner.toString() here instead of just owner

kunstmusik (approved):

PR #588, opened 2024-10-30
PE-6535: implement eth wallet support

kunstmusik:


src/components/devtools/TransferIO.tsx
```diff
@@ -42,7 +47,10 @@ function TransferIO() {
   async function confirmTransfer() {
     try {
       setTransfering(true);
-      if (isArweaveTransactionID(toAddress.trim())) {
+      if (
+        isEthAddress(toAddress.trim()) ||
```
atticusofsparta, line 51: May a single util would be good here. We may add solana and other chains in the future.

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
```diff
@@ -110,6 +131,59 @@ function ConnectWalletModal(): JSX.Element {
...
+        });
+
+        closeModal({ next: true, address });
+      } catch (error: any) {
+        eventEmitter.emit('error', error);
+      }
+    };
+
+    handleEthAccount();
+  }, [ethAccount]);
```
atticusofsparta, line 185: recommend a reducer action util for this - dispatchArNSWalletConnector or some such.

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
```diff
@@ -155,9 +225,47 @@ function ConnectWalletModal(): JSX.Element {
...
+              await viemConnector.connect();
+            } catch {
+              throw new MetamaskError('Metamask not connected');
+            } finally {
+              setConnecting(false);
+            }
+          }}
+        >
+          <MetamaskIcon className="external-icon size-12 p-3" />
+          {window.ethereum?.isMetaMask ? 'Metamask' : 'Install Metamask'}
```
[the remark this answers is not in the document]
atticusofsparta: ethereum is nullable - not a standard window object.

src/components/modals/ant-management/UpgradeAntModal/UpgradeAntModal.tsx
```diff
@@ -52,15 +52,15 @@ function UpgradeAntModal({
...
+      if (!wallet?.contractSigner || !walletAddress) {
         throw new Error('No ArConnect Signer found');
       }
       if (!data?.luaCodeTx) {
         throw new Error('No Lua Code Transaction found');
       }
       setUpgrading(true);
 
-      const signer = createAoSigner(wallet?.arconnectSigner as ContractSigner);
+      const signer = createAoSigner(wallet?.contractSigner as ContractSigner);
```
atticusofsparta, line 63: Wondering if we should have wallet.createAoSigner or wallet.createDataItemSigner - can implement the specific chain creation on each respective connector, then we can just do signer = wallet.createDataItemSigner() to get it.

src/components/pages/Settings/ArNSSettings.tsx
```diff
@@ -45,7 +45,7 @@ function ArNSSettings() {
           processId: id.trim(),
           ao: aoClient,
         }),
-        ...(wallet?.arconnectSigner ? { signer: wallet.arconnectSigner } : {}),
+        ...(wallet?.contractSigner ? { signer: wallet.contractSigner } : {}),
```
atticusofsparta, line 48: signer is supposed to be able to be undefined, whats up with this?

src/components/pages/Transaction/TransactionReview.tsx
```diff
@@ -93,6 +93,12 @@ function TransactionReview() {
     walletAddress,
   ]);
 
+  useEffect(() => {
+    if (interactionResult?.id) {
+      navigate('/transaction/complete');
+    }
+  }, [interactionResult, navigate]);
+
```
atticusofsparta, line 101: https://www.techjunkie.com/wp-content/uploads/2018/03/Sad-Gif-with-Crying-People-1.gif

src/components/pages/Transaction/TransactionReview.tsx
```diff
@@ -113,12 +119,11 @@ function TransactionReview() {
...
         processId: ioProcessId,
         dispatch: dispatchTransactionState,
-        signer: wallet?.arconnectSigner,
+        signer: wallet?.contractSigner,
         ao: aoClient,
         scheduler: aoNetwork.SCHEDULER,
       });
-
-      navigate('/transaction/complete');
+      // navigate('/transaction/complete');
```
atticusofsparta, line 126: can remove - or if not put a comment as to the previous issue with ref to the stack overflow posts.

src/services/wallets/EthWalletConnector.ts
```diff
@@ -0,0 +1,127 @@
...
+  address: EthAddress;
+  disconnectCallback: DisconnectMutateAsync<unknown>;
+  connector: Connector;
+
+  constructor(
+    ethAccount: UseAccountReturnType<Config>,
+    address: EthAddress,
+    disconnectAsync: DisconnectMutateAsync<unknown>,
+    signMessage: UseSignMessageReturnType<unknown>,
+    connector: Connector,
```
atticusofsparta, line 42: recommend moving to a configuration where you pass in the wagmi connector (the one defined in main.tsx) and use methods on that for signing and creating signers.

For reference: https://github.com/project-kardeshev/ao-wallet-kit/blob/2903b5099d10390138767e8b4f5b5ed7e9ad5e15/src/strategy/strategies/Wagmi.ts#L27

src/services/wallets/EthWalletConnector.ts
```diff
@@ -0,0 +1,127 @@
...
+  ) {
+    this.disconnectCallback = disconnectAsync;
+    this.address = address;
+    this.ethAccount = ethAccount;
+    this.connector = connector;
+
+    const provider = {
```
atticusofsparta, line 49: recommend an approach like so: https://github.com/project-kardeshev/ao-wallet-kit/blob/2903b5099d10390138767e8b4f5b5ed7e9ad5e15/src/utils/ethereum.ts#L86-L123

src/services/wallets/index.ts
```diff
@@ -1,3 +1,4 @@
 import { ArConnectWalletConnector } from './ArConnectWalletConnector';
+import { EthWalletConnector } from './EthWalletConnector';
 
-export { ArConnectWalletConnector };
+export { ArConnectWalletConnector, EthWalletConnector };
```
atticusofsparta, line 4: nice, please add ArweaveAppWalletConnector as well

src/state/contexts/WalletState.tsx
```diff
@@ -172,6 +189,39 @@ export function WalletStateProvider({
...
+
+          dispatchWalletState({
+            type: 'setWalletAddress',
+            payload: ethAccount.address,
+          });
+          dispatchWalletState({
+            type: 'setWallet',
+            payload: connector,
+          });
+        }
```
atticusofsparta, line 224: see comments on eth wallet connector re wagmi config

atticusofsparta (commented):
Looks good overall, main thing to look at I think is using functions exported by hooks in the constructor of the connector. Using the wagmi config and other utility functions is a better option since the functions exported by hooks are subject to change, where as the utilities on the config object regenerate them on each use based on the current connection state.

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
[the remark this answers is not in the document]
atticusofsparta: Should add a testing note for gisela to test with metamask installed and without metamask installed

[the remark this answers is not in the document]
kunstmusik: The suggested hange is for checking window as nullable; ethereum? is already checking if ethereum is present or not. I'll add window? but only place I could see that happening is in a unit test running in node? 

Good idea for testing note, will add a note.

src/components/pages/Settings/ArNSSettings.tsx
kunstmusik: Unsure, the code was already here that way with arconnectSigner and I modified for the field name change. Looks like there is no overload that takes in both AoProcess and a signer (there's some that uses processId + signer). I assume this works as before as all I did was change the fields, but it seems like something is off here.

src/services/wallets/EthWalletConnector.ts
```diff
@@ -0,0 +1,127 @@
...
+import { DisconnectMutateAsync } from 'wagmi/query';
+
+import {
+  AoAddress,
+  ArNSWalletConnector,
+  EthAddress,
+  WALLET_TYPES,
+} from '../../types';
+
+export const ARCONNECT_WALLET_PERMISSIONS: PermissionType[] = [
```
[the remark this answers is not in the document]
kunstmusik: Looks like copy/pasta. I removed from both ArweaveAppWallet and EthWalletConnector.

src/services/wallets/index.ts
kunstmusik: added

src/components/modals/ConnectWalletModal/ConnectWalletModal.tsx
kunstmusik: Think it may not be necessary now that I moved to wagmi config-based EthWalletConnector. (Managed to remove this block of code altogether.)

src/services/wallets/EthWalletConnector.ts
kunstmusik: Done. May have some further opportunities to refine but I think it's in good enough shape now.

src/services/wallets/EthWalletConnector.ts
kunstmusik: done

src/state/contexts/WalletState.tsx
```diff
@@ -63,18 +72,24 @@ export function WalletStateProvider({
 
   const { walletAddress, wallet } = state;
 
+  const ethAccount = useAccount();
+  const { disconnectAsync: ethDisconnect } = useDisconnect();
+  const signMessage = useSignMessage();
+  const connectors = useConnectors();
```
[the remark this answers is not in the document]
kunstmusik: done

atticusofsparta (approved):

PR #597, opened 2024-11-18
fix(PE-7135): upgrade workflow from handlers

atticusofsparta:
Converts previous workflow using Lua tx id to analyze necessary updates for ANTs to using the handlers present on the ANTs to upgrade them.


src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -119,17 +124,18 @@ const DomainsTable = ({
...
-              ant?.Controllers,
+              ant?.state?.Owner,
+              ant?.state?.Controllers,
               walletAddress?.toString(),
             ) ?? 'N/A',
           processId: record.processId,
-          targetId: ant.Records?.['@']?.transactionId ?? 'N/A',
-          sourceCode: (ant as any)?.['Source-Code-TX-ID'],
+          targetId: ant?.state?.Records?.['@']?.transactionId ?? 'N/A',
+          sourceCode: (ant as any)?.state?.['Source-Code-TX-ID'],
```
dtfiedler, line 133:
```suggestion
```

src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -114,22 +128,36 @@ const DomainsTable = ({
...
+          walletAddress && ant?.state && record?.processId
+            ? !doAntsRequireUpdate({
+                ants: {
+                  [record.processId]: {
+                    state: ant.state,
+                    handlers: ant.handlers,
+                  },
+                },
+                userAddress: walletAddress.toString(),
+              })
```
dtfiedler, line 141: this function looks like it still checks lua source code tx id vs. just validating the handlers exist on the ANT - can you confirm?

atticusofsparta: Nope - this is checking that 1, the ant can be updated by the current user, and 2, if the handlers match those listed in the SDK

dtfiedler (approved):

PR #602, opened 2024-11-26
fix(PE-7086): update sdk and changelog for ANTs

atticusofsparta:


assets/ant-changelog.md
```diff
@@ -6,6 +6,35 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 ## [Unreleased]
 
+## [8] - [XP9_LFTae8C0yvCb_DUJaC5LXaiZIbiGT1yY25X0JCg] - (2024-11-25)
+
+### Added
+
+- Release-Name Handler
+  - Calls the IO Network Process to release the specified ArNS name that is registered to the ANT.
+- Reassign-Name Handler
+  - Calls the IO Network Process to assign a new ANT Process to the respective name - must be a name registered the the ANT in question.
```
kunstmusik, line 16: fix "the the"

kunstmusik: Also a general note about the ant changelog: it's a bit technical IMO for something that is getting surfaced to end users. It could be rewritten in ways that speak that audience a bit more, but it's tricky as this is reusing the changelog that's appropriate for devs.

kunstmusik: Yes but the changelog gets shown when a lay user of ArNS sees the update button and goes to update. So it's good if you're in the ant source code repo, but not good for ArNS user. Let's go forward with this now but keep this in mind for the future that there are two audiences here.

kunstmusik (approved):

dtfiedler (approved):

PR #606, opened 2024-12-02
Remove primary name

atticusofsparta:


src/components/data-display/tables/UndernamesTable.tsx
```diff
@@ -184,53 +187,72 @@ const UndernamesTable = ({
...
-                        type: 'setTransactionData',
-                        payload: {
-                          name: targetName,
-                          ioProcessId,
-                          assetId: ioProcessId,
-                          functionName: 'primaryNameRequest',
-                        },
-                      });
-                    }
+                <Tooltip
```
kunstmusik, line 190: This code looks repeated in UndernamesSubtable. Would be nice to extract a reusable component and share between the two.

PR #609, opened 2024-12-05
fix(lua): update lua ID and changelog

atticusofsparta:


assets/ant-changelog.md
```diff
@@ -6,6 +6,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+## [9] - [ezS3Z57rq_0skoG0WYmIqJ33mJiu0HbYNn9vEu12Mc4] - (2024-12-4)
+
+### Changed
+
+- Added Ethereum address support.
+- Added Allow-Unsafe-Addresses flag to allow skipping of address validation for future compatibility with different signature algorithms on the below API's.
```
kunstmusik, line 14:
```suggestion
- Added Allow-Unsafe-Addresses flag to allow skipping of address validation for future compatibility with different signature algorithms for the following API methods:
```

atticusofsparta: This is not the place to change this, if we want to update it we should do it on the ant repo i think

kunstmusik (dismissed):

PR #612, opened 2024-12-09
POC: Optimize asset loading

atticusofsparta:


src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -58,7 +58,7 @@ type TableData = {
     supported: number;
   };
   expiryDate: string;
-  handlers: AoANTHandler[];
+  handlers: AoANTHandler[] | null;
```
dtfiedler, line 61: why `null`?

src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -288,6 +286,10 @@ const DomainsTable = ({
             );
           }
           case 'role':
+            if (loading && row.getValue(key) === 'N/A')
```
dtfiedler, line 289: why do you need the 'N/A' check?

PR #624, opened 2024-12-13
feat(PE-7068): reassign name

atticusofsparta:


src/components/modals/ant-management/ReassignNameModal/ReassignNameModal.tsx
```diff
@@ -0,0 +1,643 @@
...
+        emitter: arnsEmitter,
+        ao: aoClient,
+      });
+      handleClose();
+    } catch (error) {
+      eventEmitter.emit('error', error);
+    }
+  }
+
+  if (!show) return <></>;
```
fedellen, line 167: nit: could have this at the top of the component to prevent the above code from running

src/components/modals/ant-management/ReassignNameModal/ReassignNameModal.tsx
```diff
@@ -0,0 +1,643 @@
...
+              onClick={() => setWorkflow(REASSIGN_NAME_WORKFLOWS.NEW_EXISTING)}
+            >
+              Create new ANT
+            </button>
+            <button
+              className="bg-primary-thin p-3 rounded"
+              onClick={() => setWorkflow(REASSIGN_NAME_WORKFLOWS.EXISTING)}
+            >
+              Use existing ANT
+            </button>
```
fedellen, line 195: nit: common `<button>`s could share a component to share css styling

src/components/modals/ant-management/ReassignNameModal/ReassignNameModal.tsx
```diff
@@ -0,0 +1,643 @@
...
+  }
+
+  if (!show) return <></>;
+
+  if (!workflow) {
+    return (
+      <div className="modal-container">
+        <div className="flex flex-col rounded-md border border-dark-grey bg-foreground p-6 w-[32rem] gap-8 text-white">
+          <div className="flex flex-row justify-between items-center">
+            <h2 className="text-[24px]">Name Reassignment</h2>
```
fedellen, line 174: does `test-[24px]` become responsive by override elsewhere? I would expect a callout to a text variable that handles different screen widths

fedellen: especially since they're hard coded throughout. getting each view right for each screen widths would prove more challenging without some shared responsiveness in the app

fedellen: not necessary to do any of this in this PR's context, just calling out some nits

src/components/modals/ant-management/ReassignNameModal/ReassignNameModal.tsx
```diff
@@ -0,0 +1,643 @@
...
+                            : 'text-grey'
+                        } whitespace-nowrap cursor-pointer hover:text-white`}
+                      >
+                        Use blank slate
+                      </label>
+                    ),
+                    value: REASSIGN_NAME_WORKFLOWS.NEW_BLANK,
+                    className:
+                      'size-[25px] cursor-pointer rounded-full border  border-white shadow-[0_2px_10px] shadow-black outline-none hover:bg-white focus:shadow-[0_0_0_2px] focus:shadow-black transition-all',
+                  },
```
fedellen, line 559: nit: repeated styles, could share a simple var: 


```
const radioGroupReassignNameStyles: string = `my-styles`
```

or go more fancy, depending on how widespread styles are repeated

```
const radioGroupReassignNameStyles: string = `${commonStyles} ${radioGroupStyles} ${reassignNameStyles}`
```

src/state/actions/dispatchANTInteraction.ts
```diff
@@ -136,6 +146,30 @@ export default async function dispatchANTInteraction({
...
+        );
+        // for UX so the user sees the signing message
+        await sleep(2000);
+        result = await antProcess.reassignName({
+          name: payload.name,
+          arioProcessId: payload.arioProcessId,
+          antProcessId: newAntProcessId,
+        });
+        break;
+      }
```
fedellen, line 172: could explain further on the comment here. Also, `await spawnANT` should not end until that message is already signed

fedellen: ah, I get it -- its to see the `    'Reassigning ArNS name, please wait... ' +` message

fedellen (dismissed):
left a few nits and questions, looking good!

src/components/modals/ant-management/ReassignNameModal/ReassignNameModal.tsx
atticusofsparta: Hooks need to be mounted deterministically

src/components/modals/ant-management/ReassignNameModal/ReassignNameModal.tsx
atticusofsparta: Good point, i should be using rem here

src/state/actions/dispatchANTInteraction.ts
atticusofsparta: Exactly yeah - noticed the reassign was a bit quick sometimes and it didn't even give the message time to rerender.

src/components/modals/ant-management/ReassignNameModal/ReassignNameModal.tsx
atticusofsparta: That being said, because this uses the useDomainInfo hook, it *shouldnt* be doing anything extra here, since it is being access from a page that is also using that hook - internally it uses useQuery which will dedupe the request and return the same data to all the components calling it.

src/components/pages/ManageDomain/ManageDomain.tsx
```diff
@@ -119,44 +122,48 @@ function ManageDomain() {
...
-            <Star className={`w-[16px]`} />{' '}
-            {name == primaryNameData?.name ? 'Remove Primary' : 'Make Primary'}
-          </button>
+                dispatchModalState({
+                  type: 'setModalOpen',
+                  payload: { showPrimaryNameModal: true },
+                });
+              }}
+            >
+              <Star className={`w-[16px]`} />{' '}
```
kunstmusik, line 161: use rem

kunstmusik (approved):

PR #625, opened 2024-12-13
fix(PE-7306): allow access to extend lease page

atticusofsparta:


src/components/layout/ExtendLease/ExtendLease.tsx
```diff
@@ -434,7 +431,7 @@ function ExtendLease() {
               <></>
             )
           }
-        />
+        />{' '}
```
dtfiedler, line 434:
```suggestion
        />
```

src/hooks/useDomainInfo.tsx
```diff
@@ -51,6 +51,7 @@ export default function useDomainInfo({
     queryKey: ['domainInfo', { domain, antId, arioProcessId, aoNetwork }],
     queryFn: () => getDomainInfo({ domain, antId }).catch((error) => error),
     refetchOnWindowFocus: false,
+    staleTime: Infinity,
```
dtfiedler, line 54: this might be better as 1 hour given names can/will be reassigned/released

dtfiedler (dismissed):

src/hooks/useDomainInfo.tsx
atticusofsparta: When we reassign/release locally we reset this query - going to go ahead and send the to QA but will be sure to add this to the improvement to be done on this hook. Would like to break up what its doing on some of the info to separate queries.

PR #626, opened 2024-12-13
fix(PE-6993): add ANT cu url on all ant usage

atticusofsparta:


dtfiedler (approved):

src/utils/constants.ts
```diff
@@ -112,6 +112,7 @@ export const NETWORK_DEFAULTS = {
       CU_URL: ANT_AO_CU_URL,
       MU_URL: 'https://mu.ao-testnet.xyz',
       SCHEDULER: DEFAULT_SCHEDULER_ID,
+      GRAPHQL_URL: ARWEAVE_GRAPHQL_URL,
```
dtfiedler, line 115: what's the reason we are not deferring to AO defaults for MU, SCHEDULER and GRAPHQL?

dtfiedler (approved):

src/utils/constants.ts
atticusofsparta: Mostly for easy of control

dtfiedler: have we ever used it?

package.json
```diff
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "@ant-design/icons": "5.4.0",
-    "@ar.io/sdk": "^3.3.0-alpha.11",
+    "@ar.io/sdk": "^3.3.0-alpha.15",
```
dtfiedler, line 30: 3.3.0

dtfiedler (changes requested):
rebase and update sdk

src/utils/constants.ts
atticusofsparta: I believe its technically used whenever you spawn a process, so for example we may want to point it at goldsky instead of arweave.net. But no I don't think we use it much in normal operations, thats mainly the MU, and CU

package.json
atticusofsparta: https://github.com/ar-io/arns-react/pull/626/commits/139b72b87773a8d4a790dbb35ab4cfac1f9e0476

dtfiedler (approved):

PR #627, opened 2024-12-16
fix(PE-7234): add arns domain resolver

atticusofsparta:


src/components/cards/ARNSCard/ARNSCard.tsx
```diff
@@ -8,7 +8,7 @@ const protocol = 'https';
 
 function ARNSCard({
   domain,
-  gateway = ARWEAVE_HOST,
+  gateway = NETWORK_DEFAULTS.ARNS.RESOLVER,
```
dtfiedler, line 11: nit: this is pointing to an `AR-IO` gateway, and not it's resolver endpoint. might be better named as `NETWORK_DEFAULTS.ARIO.GATEWAY` or similar

atticusofsparta: Decided to go with "HOST" instead since gateway is a loaded term in our codebase [78634f4](https://github.com/ar-io/arns-react/pull/627/commits/78634f49f9778e42147d60f542a7bde06f488f63)

dtfiedler (approved):

PR #628, opened 2024-12-17
fix(nav menu): add ant logo associate with primary name to nav menu

atticusofsparta:


src/components/cards/NavMenuCard/NavMenuCard.tsx
```diff
@@ -371,36 +377,48 @@ function NavMenuCard() {
...
-            />
-          )}
+          <span className="flex text-white gap-2 items-center justify-center text-[14px] h-[2.5rem] p-2">
+            {primaryNameData?.name ? (
+              <>
+                <AntLogoIcon
+                  className="rounded-full max-h-[1.875rem] border border-dark-grey"
+                  id={domainDomain?.info?.Logo}
+                  icon={
+                    <div className="flex items-center justify-center rounded-full fill-white p-[5px]">
```
kunstmusik, line 405: In general need to stop using px and use rem by default. This will need to get fixed during audit.

kunstmusik (approved):

PR #639, opened 2025-01-09
feat(PE-7365): rnp page

atticusofsparta:


src/hooks/useCostDetails.tsx
```diff
@@ -1,15 +1,33 @@
...
+  CostDetailsResult,
+} from '@ar.io/sdk';
 import { useGlobalState } from '@src/state';
 import { useQuery } from '@tanstack/react-query';
 
-export function useCostDetails(params: AoGetCostDetailsParams) {
-  const [{ arioProcessId, arioContract }] = useGlobalState();
-  // we are verbose here to enable predictable keys. Passing in the entire params as a single object can have unpredictable side effects
-  return useQuery({
+export function buildCostDetailsQuery(
```
kunstmusik, line 10: Not sure why this function is separated out, would re-inline this code back into useCostDetails

src/components/data-display/tables/TableView.tsx
```diff
@@ -96,10 +96,7 @@ const TableView = <T, S>({
     <>
       <div className="overflow-x-auto scrollbar">
         <table className={'w-full table-auto' + ' ' + tableClass}>
-          <thead
-            className={`text-[14px] text-grey ${headerClass}`}
-            style={{ borderRadius: '10px' }}
-          >
+          <thead className={`text-[14px] text-grey ${headerClass}`}>
```
kunstmusik, line 99: use rem

src/components/data-display/charts/RNPChart.tsx
```diff
@@ -180,9 +183,20 @@ export function RNPChart({
...
+        <Lottie animationData={arioLoading} loop={true} className="h-[100px]" />
+        <span>Loading Price Chart...</span>
+      </span>
+    );
 
-  if (!costDetails?.returnedNameDetails) return <></>;
+  if (!costDetails?.returnedNameDetails)
+    return (
+      <div className="flex text-white text-lg justify-center items-center m-auto">
+        This is not a returned name, unable to load chart.
```
kunstmusik, line 197: Odd, seems like if this is the case, this shouldn't really be shown and error handled elsewhere. Could return empty <></> for example here.

src/components/inputs/Switch.tsx
```diff
@@ -0,0 +1,33 @@
...
+    root?: string;
+    thumb?: string;
+  };
+}) => (
+  <RadixSwitch.Root
+    onCheckedChange={onChange}
+    checked={checked}
+    className={
+      className?.root ??
+      'relative h-[25px] w-[42px] cursor-default rounded-full bg-blackA6 shadow-[0_2px_10px] shadow-blackA4 outline-none focus:shadow-[0_0_0_2px] focus:shadow-black data-[state=checked]:bg-black'
```
kunstmusik, line 21: Would be good to use rem's here too.

src/components/data-display/tables/ReturnedNamesTable.tsx
```diff
@@ -0,0 +1,484 @@
...
+
+  useEffect(() => {
+    if (filter) {
+      setFilteredTableData(filterTableData(filter, tableData));
+    } else {
+      setFilteredTableData([]);
+    }
+  }, [filter, tableData]);
+  // Define columns for the table
+  const columns: ColumnDef<TableData, any>[] = [
```
kunstmusik, line 219: I find this mapping here to be less clear vs. just using code for each column def entry since you have lots of conditionals below. Not much saved and harder to read.

src/components/data-display/tables/ReturnedNamesTable.tsx
```diff
@@ -0,0 +1,484 @@
...
+                  )}
+                  <button
+                    className="p-2 py-[0.4rem] text-center size-fit rounded text-black bg-primary"
+                    onClick={() => {
+                      navigate(
+                        `/register/${lowerCaseDomain(row.original.name)}`,
+                      );
+                    }}
+                  >
+                    Buy
```
kunstmusik, line 355: Odd this is Buy even if you're going to lease. Seems like this should be "Purchase" or "Register" (It's "Register" on the main page with new names.)

src/hooks/useCostDetails.tsx
atticusofsparta: I pulled this out now because we are using the cost details api on the table to load prices and want to leverage the query cache - seperating the function like this ensures we use the same query args (key, function, staletime primaryily)

src/components/data-display/tables/ReturnedNamesTable.tsx
atticusofsparta: Register would be better, yeah

atticusofsparta: https://github.com/ar-io/arns-react/pull/639/commits/fde86082de9ce23903f841989728d0751f3ecb7f

src/components/data-display/tables/ReturnedNamesTable.tsx
atticusofsparta: I like it, its nice to have all the headers at the top for making new tables. I can just copy paste the config and then edit them all in one place.

src/components/inputs/Switch.tsx
atticusofsparta: [fb6ff4c](https://github.com/ar-io/arns-react/pull/639/commits/fb6ff4cb2ca616122b7253290564f3c2f3f54533)

src/components/data-display/tables/ReturnedNamesTable.tsx
kunstmusik: Not a fan of the map. Can copy/paste configs without using map.

src/components/data-display/charts/RNPChart.tsx
atticusofsparta: Yeah in theory this should never be reached 🤔 ok will just do a fragment

src/components/data-display/tables/TableView.tsx
atticusofsparta: https://github.com/ar-io/arns-react/pull/639/commits/63b4070b3fe3667806bc0fb9d67268fd28f1a256

src/components/data-display/tables/ReturnedNamesTable.tsx
atticusofsparta: Well there 3-4 other tables that have this setup, so ill make a ticket for a refactor as tech debt

kunstmusik (approved):

PR #651, opened 2025-01-27
Release single undername fix.

atticusofsparta:
Note: this will not fix current ANTs because the upgrade checks do not know how to see the change. We will be releasing the altered upgrade workflow to handle this later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "@ant-design/icons": "5.4.0",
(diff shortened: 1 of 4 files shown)


dtfiedler (approved):
the scope was missed to address existing ANT holders. so this ultimately doesn't solve those problems. Specifically, ar-io-nodes are unable to resolve these names - creating a bad UX. 

while this is fine to ship as-is to prevent new ANTs from the behavior, it's important to remember why we created the defect in the first place.

PR #666, opened 2025-02-07
fix(PE-7533): asset loading improvements

atticusofsparta:


src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -288,6 +286,10 @@ const DomainsTable = ({
             );
           }
           case 'role':
+            if (loading && row.getValue(key) === 'N/A')
```
kunstmusik, line 289: could do something like:

```
return (loading && row.getValue(key) === 'N/A') ? (<span>...</span>) : capitalize(row.getValue(key));
````

File: src/state/actions/dispatchArNSUpdate.ts [hunk 2]
```diff
@@ -3,140 +3,214 @@ import {
...
+      process: new AOProcess({ processId: arioProcessId, ao }),
+    }) as AoARIORead;
+    const antRegistry = ANTRegistry.init({
+      process: new AOProcess({
+        processId: ANT_REGISTRY_ID,
+        ao: connect(aoNetworkSettings.ANT),
+      }),
+    });
+    const [arnsRecords, allUserAnts] = await Promise.all([
+      arioContract.getArNSRecords({ limit: 100_000 }),
```
kunstmusik, line 67: So if we have more than 100K records this won't retrieve all of them? (And if we have that many this whole scheme has bigger problems)

src/state/actions/dispatchArNSUpdate.ts
```diff
@@ -3,140 +3,214 @@ import {
...
-            return msg.Tags.find((t: Tag) => t.name === 'Error');
-          }) !== undefined;
+    const flatAntIds = new Set([
+      ...allUserAnts.Controlled,
+      ...allUserAnts.Owned,
+    ]);
 
-        if (hasError) {
-          return [] as AoANTHandler[];
+    const userDomains = arnsRecords.items.reduce(
```
kunstmusik, line 76: In general the scheme to download all records then filter isn't going to scale. We need a different path moving forward, perhaps adding something to the SDK/contract to allow querying passed-in list of ids or something of that sort.

src/state/actions/dispatchArNSUpdate.ts
kunstmusik: FWIW: looks like with testnet it's a 192kb download each app start/refresh of table

kunstmusik (dismissed):

src/state/actions/dispatchArNSUpdate.ts
dtfiedler: agree we'll want something like this - but will punt it to post launch

dtfiedler (dismissed):
🚢

src/state/actions/dispatchArNSUpdate.ts
dtfiedler, line 67: i'd consider paginating over multiple requests, say every 1000 to avoid the size of this payload taking a long time to serve - add a TODO that this may hit limitations

dtfiedler: you can also tell the table to only fetch the "latest" records by storing the last cursor in local storage, and on reload populate it from that. name assignments will happen infrequently, and initial load should catch them. this would allow you to cache previously loaded names and only update for new ones.

atticusofsparta: yes - this will be reduced significantly with the filter by process ID changes

src/components/data-display/tables/TableView.tsx
```diff
@@ -227,20 +228,116 @@ const TableView = <T, S>({
...
-                  i === table.getState().pagination.pageIndex
-                    ? 'bg-dark-grey text-white'
-                    : 'text-grey'
-                }`}
-              >
-                {i + 1}
-              </button>
-            ))}
+
+            {table.getPageCount() <= 6 ? (
```
kunstmusik, line 232: lot of logic for page selector control, would be good to refactor to a separate component in the future.

src/components/forms/DomainSettings/DomainSettings.tsx
```diff
@@ -113,22 +113,12 @@ function DomainSettings({
...
-      if (data?.processId) {
-        queryClient.invalidateQueries({
-          queryKey: ['ant', data?.processId.toString()],
-          refetchType: 'all',
-        });
-      }
-
       queryClient.invalidateQueries({
-        queryKey: ['domainInfo'],
+        queryKey: ['domainInfo', data?.processId.toString(), domain],
```
kunstmusik, line 117: fix this cache invalidation

atticusofsparta: [45a10f5](https://github.com/ar-io/arns-react/pull/666/commits/45a10f59a833ec0ccb4352db4cf412c7adbd8c87)

kunstmusik (approved):

PR #676, opened 2025-02-12
fix(PE-7658): fix buy name

atticusofsparta:


dtfiedler (commented):
Bump the SDK to 3.5.3 too

dtfiedler (approved):

PR #681, opened 2025-02-12
fix(PE-7652): make update button only display for owners

atticusofsparta:


kunstmusik:
@atticusofsparta needs a changelog entry

kunstmusik (approved):

PR #699, opened 2025-02-25
feat(PE-7728): adds prices page on a /prices route

atticusofsparta:


src/components/data-display/tables/TableView.tsx
```diff
@@ -102,8 +104,8 @@ const TableView = <T, S>({
   return (
     <>
       <div className={`overflow-x-auto scrollbar ${tableWrapperClassName}`}>
-        <table className={'w-full table-auto' + ' ' + tableClass}>
-          <thead className={`text-[0.875rem] text-grey ${headerClass}`}>
+        <table className={`${tableClass} w-full table-auto`}>
```
kunstmusik, line 107: This changes the order of classes from appending to prepending. Need to be careful here and the line below. Would suggest reordering to make it come at end to match prior behavior to limit scope of change.

src/components/data-display/tables/TableView.tsx
```diff
@@ -161,7 +163,7 @@ const TableView = <T, S>({
               </tr>
             ))}
           </thead>
-          <tbody className="text-sm">
+          <tbody className={`${bodyClass} text-sm`}>
```
kunstmusik, line 166: same here

src/components/pages/Prices/Prices.tsx
```diff
@@ -0,0 +1,159 @@
...
+  ...minFees,
+};
+
+export default function Prices() {
+  const { data: demandFactor, isLoading, error } = useDemandFactor();
+  const [domain, setDomain] = useState('');
+
+  const [tableData, setTableData] = useState<TableData[]>([]);
+
+  useEffect(() => {
```
kunstmusik, line 58: could use useMemo here

src/components/pages/Prices/Prices.tsx
```diff
@@ -0,0 +1,159 @@
...
+              eventEmitter.emit('error', {
+                name: 'Invalid ARNS name',
+                message: `${trimmed} is not a valid arns name`,
+              });
+              return;
+            }
+            setDomain(trimmed);
+          }}
+        />
+        <span className="flex p-2 bg-foreground border border-dark-grey rounded text-white whitespace-nowrap">
```
kunstmusik, line 122: Nit: this looks like a button to press

src/components/pages/Prices/Prices.tsx
```diff
@@ -0,0 +1,159 @@
...
+        const numVal = Number(row.getValue(key));
+        if (key === 'characters') return numVal.toString();
+        return formatARIOWithCommas(Number(numVal.toFixed(3)));
+      },
+    });
+  });
+
+  return (
+    <div className="page">
+      <div className="flex w-full py-2 justify-between  items-center gap-5">
```
kunstmusik, line 103: wouldn't mind seeing a little more space betwen this header row and the table (like an mb-2)

src/components/data-display/tables/TableView.tsx
atticusofsparta: yup this was a deliberate change. Prepending ensures overridability of the default classes but will still add them if not conflicting.

src/components/pages/Prices/Prices.tsx
dtfiedler: agree

src/components/pages/Prices/Prices.tsx
```diff
@@ -0,0 +1,164 @@
...
+        return data;
+      },
+    );
+    setTableData(newData);
+  }, [demandFactor, isLoading, error]);
+
+  const columns = [
+    'characters',
+    // 'baseFee',
+    //  'adjustedRegistrationFee',
```
dtfiedler, line 83: remove

src/components/data-display/tables/TableView.tsx
kunstmusik: Hmm, seems like ordering doesn't determine overridability. We may want to look at using tailwind-merge at some point. (In another PR...)

src/components/pages/Prices/Prices.tsx
atticusofsparta: https://github.com/ar-io/arns-react/pull/699/commits/53f445ffc6006961a2f0e885c61c8ec7d97a718e

src/components/data-display/tables/TableView.tsx
atticusofsparta: yup this was a deliberate change. Prepending ensures overridability of the default classes but will still add them if not conflicting.

src/components/pages/Prices/Prices.tsx
atticusofsparta: Validated all styles with jon - this isn't a user-facing UI so we can come back for any more style nits later.

dtfiedler (approved):

PR #704, opened 2025-02-28
fix(PE-7424): change upgrade workflow to reassign

atticusofsparta:
## Scope

- Adds arweave-graphql dep for querying arweave gateway for AO (ant) process headers
- - Asset resolution (`dispatchArNSUpdate`) updated to get process headers
- - those headers are used to compare the current module ID exported from the SDK to predicate necessity to upgrade and utility functions have been updated to support this
- dispatchANTInteraction updated with new workflow to upgrade ANTs (reassign to evolve workflow)
- - Spawns new ant with old ant state, reassigns target domain to new ant from old ant, and validates the domain was appropriately reassigned.
- - Props updated to provide a promise callback per step, which allows for overriding the default signing behaviour and allowing for promise based user input (such as cancellation of the workflow)
- - Starting to perform cache management in this helper, pulling out that business logic from UI components
- To facilitate this, upgrade modals have been updated to resolve the ANT id by domain name

### Misc
- Added tanstack devtools activation button to devtools modal
- Disabled unit tests


src/state/actions/dispatchANTInteraction.ts
```diff
@@ -1,36 +1,54 @@
...
   signer,
   owner,
-  dispatch,
+  dispatchTransactionState,
+  dispatchArNSState,
   ao,
+  // this can allow for waiting on promise resolution for UI input on individual steps
+  stepCallback = async (step) => {
+    if (typeof step === 'string') {
+      dispatchTransactionState({
```
kunstmusik, line 36: Unsure whether this gets the dispatchTransactionState from args here. If it does, it's still not as clear to have that as part of default arg here vs. creating the callback within the function body as it was being done before where it was creating dispatchSigningMessage. Would prefer the stepCallback function be overridden if undefined within the body.

src/utils/searchUtils/searchUtils.ts
```diff
@@ -133,17 +134,19 @@ export function getAntsRequiringUpdate({
...
 }): string[] {
   return Object.entries(ants)
     .map(([id, ant]) => {
       // if user is not the owner, skip
       if (!ant.state?.Owner || ant?.state.Owner !== userAddress) return;
-      if (!AntHandlerNames.every((handler) => ant.handlers?.includes(handler)))
+      if (
+        !ant.processMeta ||
+        ant.processMeta.tags.find(
+          (t) => t.name === 'Module' && t.value !== ANT_MODULE_ID,
```
kunstmusik, line 147: Needs to be updated for semantic versioning

src/state/actions/dispatchANTInteraction.ts
atticusofsparta: I moved it but i dont like it https://github.com/ar-io/arns-react/pull/704/commits/9de91f09f26cd63ca5be8528fc258dccf0dc4080

atticusofsparta:
to discuss: presentation of domains when unable to fetch process metadata

src/state/actions/dispatchANTInteraction.ts
```diff
@@ -172,49 +183,139 @@ export default async function dispatchANTInteraction({
...
+          process: new AOProcess({
+            processId: payload.arioProcessId,
+            ao,
+          }),
+        });
+        // poll for update on registry to update record
+        let record = undefined;
+        let retries = 0;
+        const maxRetries = 10;
+        while (record?.processId !== newAntId && retries < maxRetries) {
```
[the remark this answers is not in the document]
atticusofsparta: Implemented the concurrent pattern here https://github.com/ar-io/arns-react/pull/704/commits/459c0a92362593ab1ca4e4ed1e958ad23b5a6b8b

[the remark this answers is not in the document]
atticusofsparta: sleep moved here https://github.com/ar-io/arns-react/pull/704/commits/65bccb2e43cd5258090c451cc242a504dba2ecaa

atticusofsparta:
remove handlers usage and remove Info calls

src/utils/searchUtils/searchUtils.ts
atticusofsparta: updated this to provide a module id that is set to our state

src/state/actions/dispatchANTInteraction.ts
```diff
@@ -50,6 +43,15 @@ export default async function dispatchANTInteraction({
   ao: AoClient;
   stepCallback?: (step?: Record<string, string> | string) => Promise<void>;
 }): Promise<ContractInteraction> {
+  stepCallback ??= async (step) => {
```
dtfiedler, line 45: does this need to be `async`?

atticusofsparta: To satisfy the type, technically no, i could manually construct and return a promise, but making it async is an easier shorthand.

The reason its async is to allow the callback to be suspended and another operation to be performed/predicated by the calling workflow - such as UI input.

PR #716, opened 2025-03-13
fix(spawn): add the logo on ant spawns

atticusofsparta:


.husky/pre-push
```diff
@@ -2,5 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 # yarn test
-yarn build
-yarn validate-ant-changelog
\ No newline at end of file
+yarn build
```
kunstmusik, line 5: why was yarn validate-ant-changelog removed?

atticusofsparta: Because we are going to be shifting toward a different changelog pattern, but will handle adding that in another ticket

kunstmusik (approved):

dtfiedler (approved):

PR #722, opened 2025-03-21
feat(PE-7852): add turbo credit balance to nav card

atticusofsparta:
Verify the target branch before merging - need to discuss if this is an all-at-once feature or iterative

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -122,6 +122,11 @@ export const NETWORK_DEFAULTS = {
   ARNS: {
     HOST: 'ar.io',
   },
+  TURBO: {
+    UPLOAD_URL: 'https://turbo.ardrive.io',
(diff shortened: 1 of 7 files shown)


src/utils/constants.ts
```diff
@@ -122,6 +122,11 @@ export const NETWORK_DEFAULTS = {
   ARNS: {
     HOST: 'ar.io',
   },
+  TURBO: {
+    UPLOAD_URL: 'https://turbo.ardrive.io',
```
fedellen, line 126: turbo works but will forward traffic, a more direct route`https://upload.ardrive.io`

fedellen (dismissed):

dtfiedler:
i'd rec releasing this as-is before bolting on any turbo integrated features

PR #723, opened 2025-03-22
fix(PE-7871): create sentry logger and pass to ario processes

atticusofsparta:


src/hooks/useDomainInfo.tsx
```diff
@@ -110,6 +111,10 @@ export function buildDomainInfoQuery({
         process: new AOProcess({
           processId,
           ao: antAo,
+          logger: new SentryLogger({
+            component: 'ANT',
```
arielmelendez, line 115: Why ANT as opposed to something like ArNS-App?

arielmelendez: I read more about how these are used in Sentry and I think you're free to choose these however is meaningful for grouping within the project. ANT is fine by me if you think it makes sense organizationally.

src/utils/logger.ts
```diff
@@ -0,0 +1,70 @@
...
+    logLevel?: LogLevel;
+  }) {
+    super({ level: logLevel });
+    this.component = component;
+    this.context = initialContext;
+    this.logLevel = logLevel;
+  }
+
+  setContext(context: LogContext): SentryLogger {
+    this.context = { ...this.context, ...context };
```
arielmelendez, line 28: Wouldn't spreading properties this way overwrite the common ones of the first's with the second's?

arielmelendez: Accordingly, consider renaming the function `append/add/mergeContext` since `set` sounds like remove and replace to me.

src/utils/logger.ts
```diff
@@ -0,0 +1,70 @@
...
+    return this;
+  }
+
+  withContext(context: LogContext): any {
```
arielmelendez, line 32: This is reminiscent of winston logger's `child` method. Consider renaming to that in order to honor a familiar pattern in our other codebases.

src/utils/logger.ts
```diff
@@ -0,0 +1,70 @@
...
+    return new SentryLogger({
+      component: this.component,
+      initialContext: { ...this.context, ...context },
+      logLevel: this.logLevel,
+    });
+  }
+
+  warn(message: string, context?: LogContext): void {
+    this.warn(message, context);
```
arielmelendez, line 41: Shouldn't this be `super.warn` otherwise you'd have an infinite recursion?!

arielmelendez: Same goes for `error` function.

arielmelendez (changes requested):
A few questions and concerns to address.

src/utils/logger.ts
atticusofsparta: actually yes - i shouldnt be extending logger here but rather consuming a logger and implemented the logger interface and wrapping the requisite functions

PR #729, opened 2025-03-27
Release sentry http client activation

atticusofsparta:


src/utils/sentry.ts
```diff
@@ -31,6 +31,18 @@ const sentry =
               matchRoutes,
             ),
           }),
+          Sentry.httpClientIntegration({
+            failedRequestStatusCodes: [
+              [400, 499],
+              [500, 599],
```
dtfiedler, line 37: Wouldn't this be 400-599?

dtfiedler (approved):

src/utils/sentry.ts
atticusofsparta: Yes exactly - i supposed i could have used a single range specifier instead of 2

PR #757, opened 2025-05-03
fix(checkout): fix form validation

atticusofsparta:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved form validation for payment options, ensuring the form is only valid when there is sufficient balance for credits and crypto payment methods.
  - Adjusted how form validity is managed when switching between payment methods to prevent unintended validation issues.

- **Chores**
  - Updated a package import source for improved dependency management.

### DIFF
--- a/src/components/forms/PaymentOptionsForm/PaymentOptionsForm.tsx
+++ b/src/components/forms/PaymentOptionsForm/PaymentOptionsForm.tsx
@@ -373,6 +373,12 @@ function PaymentOptionsForm({
 
   const [showTopupModal, setShowTopupModal] = useState(false);
 
+  useEffect(() => {
+    if (paymentMethod === 'credits' || paymentMethod === 'crypto') {
+      setIsValid(!isInsufficientBalance);
+    }
+  }, [paymentMethod, isInsufficientBalance]);
+
   return (
     <>
       <div className="flex flex-col gap-6 w-full">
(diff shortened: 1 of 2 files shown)


src/components/forms/PaymentOptionsForm/PaymentOptionsForm.tsx
```diff
@@ -423,7 +429,7 @@ function PaymentOptionsForm({
             className={`flex flex-col data-[state=active]:p-6 data-[state=active]:border border-dark-grey rounded h-full data-[state=inactive]:size-0 data-[state=inactive]:opacity-0 data-[state=active]:min-h-[405px]`}
           >
             <CardPanel
-              setIsValid={setIsValid}
+              setIsValid={paymentMethod === 'card' ? setIsValid : () => null}
```
dtfiedler, line 432: nit: add this logic to to useEffect vs here. 


i.e. if paymentMethod === "card" setIsValid(true)

atticusofsparta: I think lets leave that for now - the organization of the callback is a bit wierd because setIsValid is passed into CardPanel and the validation there is handled with callbacks from internal components, but theres no panel components for the other methods. So we dont want to always isValid to true on that panel.

PR #800, opened 2025-06-27
fix(manage): reduce use of dispatchArNSState

dtfiedler:


src/components/pages/Manage/Manage.tsx
```diff
@@ -1,29 +1,53 @@
...
+  const { ants: antsRequireUpdate, isLoading: isLoadingAntsRequireUpdate } =
+    useAntsRequireUpdate();
+  const [isLoading, setIsLoading] = useState(isLoadingDomains || isLoadingAnts);
+
+  useEffect(() => {
+    console.log('isLoadingDomains', isLoadingDomains);
+    console.log('isLoadingAnts', isLoadingAnts);
+    console.log('isRefetchingDomains', isRefetchingDomains);
+    console.log('isRefetchingAnts', isRefetchingAnts);
+    console.log('isLoadingAntsRequireUpdate', isLoadingAntsRequireUpdate);
```
dtfiedler, line 40:
```suggestion
```

src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -147,22 +140,7 @@ const DomainsTable = ({
...
-                    version: ant.version,
-                    processMeta: ant.processMeta,
-                  },
-                },
-                userAddress: walletAddress.toString(),
-                currentModuleId: antModuleId,
-              })
-            : false);
-
+        console.log(domainData);
```
atticusofsparta, line 143:
```suggestion
```

src/components/pages/Manage/Manage.tsx
```diff
@@ -1,29 +1,48 @@
...
+    setIsLoading(
+      isLoadingDomains || isLoadingAnts || isLoadingAntsRequireUpdate,
+    );
+  }, [
+    isLoadingDomains,
+    isLoadingAnts,
+    isRefetchingDomains,
+    isRefetchingAnts,
+    isLoadingAntsRequireUpdate,
+  ]);
```
atticusofsparta, line 45: useMemo is more concise here

src/hooks/useAntsForWallet.tsx
```diff
@@ -0,0 +1,115 @@
...
+        return +antData[processId].version < +(latestAntVersion?.version || 0);
+      })
+      .map((processId) => processId),
+    isLoading:
+      isLoading ||
+      isLoadingLatestAntVersion ||
+      isRefetching ||
+      isRefetchingLatestAntVersion,
+  };
+};
```
atticusofsparta, line 115: recommend useMemo on the ants and isLoading vars

deps for ants is the query data (antData and latestAntVersion) then deps for isLoading is all the loading variables.

This helps prevent unnecessary rerenders, especially if using this in a lot of elements (eg domains table on every single row and in ternaries for feature blocking inputs)

src/components/pages/Manage/Manage.tsx
```diff
@@ -1,29 +1,48 @@
...
+    isRefetching: isRefetchingAnts,
+    refetch: refetchAnts,
+  } = useAntsForWallet();
+  const {
+    data: domains = {},
+    isLoading: isLoadingDomains,
+    isRefetching: isRefetchingDomains,
+    refetch: refetchDomains,
+  } = useArNSRecordsForWallet();
+  const { ants: antsRequireUpdate, isLoading: isLoadingAntsRequireUpdate } =
```
atticusofsparta, line 31: nit: antsRequireUpdate seems to indicate boolean but this is an array of process id's - `antsRequiringUpdate` might be higher signal here.

src/hooks/useArNSRecordsForWallet.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { useGlobalState, useWalletState } from '@src/state';
+import { UseQueryResult, useQuery } from '@tanstack/react-query';
+
+export const useArNSRecordsForWallet = (): UseQueryResult<
+  Record<string, AoArNSNameData>
+> => {
+  const [{ walletAddress }] = useWalletState();
+  const [{ arioContract }] = useGlobalState();
+  return useQuery({
+    queryKey: ['arns-records-for-wallet', walletAddress?.toString()],
```
atticusofsparta, line 11: arioProcessId, aoNetwork.ARIO, and hyperbeamUrl should be added to the query key here

src/hooks/useArNSRecordsForWallet.tsx
```diff
@@ -0,0 +1,38 @@
...
+      const arnsRecordsForWallet: Record<string, AoArNSNameData> = {};
+      let hasMore = true;
+      while (hasMore) {
+        const {
+          items: newRecords,
+          hasMore: nextHasMore,
+          nextCursor,
+        } = await arioContract.getArNSRecordsForAddress({
+          address: walletAddress?.toString() ?? '',
+          cursor,
```
atticusofsparta, line 23: this this needs limit:1000 otherwise it will default to 100 right?

src/hooks/useArNSRecordsForWallet.tsx
```diff
@@ -0,0 +1,38 @@
...
+        });
+        if (!nextHasMore) {
+          break;
+        }
+        hasMore = nextHasMore;
+        cursor = nextCursor;
+      }
+      return arnsRecordsForWallet;
+    },
+    staleTime: 60 * 60 * 1000, // 1 hour
```
atticusofsparta, line 36: enabled: !!walletAddress

src/hooks/useArNSRecordsForWallet.tsx
```diff
@@ -0,0 +1,38 @@
...
+      let cursor: string | undefined = undefined;
+      const arnsRecordsForWallet: Record<string, AoArNSNameData> = {};
+      let hasMore = true;
+      while (hasMore) {
+        const {
+          items: newRecords,
+          hasMore: nextHasMore,
+          nextCursor,
+        } = await arioContract.getArNSRecordsForAddress({
+          address: walletAddress?.toString() ?? '',
```
atticusofsparta, line 22: nit: explicitly throw if no wallet address to satisfy type (also see comment on enabling the query)

src/hooks/useArNSRecordsForWallet.tsx
```diff
@@ -0,0 +1,38 @@
...
+          cursor,
+        });
```
atticusofsparta, line 24: nit: explicitly throw if no wallet address to satisfy type (also see comment on enabling the query)

Also, think we need to specify limit: 1000 else it will default to 100?

src/state/actions/dispatchArNSUpdate.ts
```diff
@@ -52,7 +52,6 @@ export async function dispatchArNSUpdate({
 
     // TODO: move all of these async calls to unique hooks and remove the need for the dispatchArNSUpdate action
     const arioContract = ARIO.init({
-      hyperbeamUrl,
```
atticusofsparta, line 55: this was added so the ANTRegistry internally in the sdk uses the hyperbeam ACL - why does it need to be removed?

src/state/contexts/ArNSState.tsx
```diff
@@ -20,6 +20,7 @@ export type ANTProcessData = {
   version: number;
   processMeta: TransactionEdge['node'] | null;
   errors?: Error[];
+  isLatestVersion?: boolean;
```
atticusofsparta, line 23: is the intention of nullish to indicate to not display upgrade prompts?

src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -346,14 +315,9 @@ const DomainsTable = ({
...
-                    dispatchANTUpdate({
-                      processId,
-                      domain: lowerCaseDomain(row.original.name),
-                      aoNetwork,
-                      queryClient,
-                      walletAddress,
-                      hyperbeamUrl,
-                      dispatch: dispatchArNSState,
+                    await queryClient.invalidateQueries({
+                      queryKey: ['ant', processId],
```
atticusofsparta, line 319: This is a less dependant way to indicate the query key - omitting 'ant' will also refresh _any_ apis using the process id.

Query key is position sensitive, so if processId is changed in the future to another index this breaks. Using the predicate helps future proof a bit there.

```suggestion
                    await queryClient.invalidateQueries({
                      predicate: ({ queryKey }) => queryKey.includes("ant") && queryKey.includes(processId),
```

atticusofsparta: Another note there, check in with the new hooks on things relying on state... I believe the useAntsForWallet hook calls this in its query, but its state may not be updated in this refresh.

Thats a case where including the process id in the query key and then predicating on the process ID can help in cache control.

src/hooks/useAntsForWallet.tsx
```diff
@@ -0,0 +1,115 @@
...
+> => {
+  const [{ aoNetwork, hyperbeamUrl }] = useGlobalState();
+  const queryClient = useQueryClient();
+  const [{ walletAddress }] = useWalletState();
+  const { data: accessControlList = { Owned: [], Controlled: [] } } =
+    useAccessControlList();
+  const { data: antVersions = {} } = useANTVersions();
+  const { data: latestAntVersion } = useLatestANTVersion();
+  return useQuery({
+    queryKey: ['ants-with-metadata', walletAddress?.toString()],
```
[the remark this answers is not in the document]
atticusofsparta: And actually doesn't look like we need wallet address here. Fine to have it, but it doesn't affect the query.

src/hooks/useAntsForWallet.tsx
```diff
@@ -0,0 +1,115 @@
...
+              isLatestVersion,
+              errors: [],
+            },
+          };
+        }),
+      );
+
+      return antsWithMetadata.reduce((acc, item) => ({ ...acc, ...item }), {});
+    },
+  });
```
atticusofsparta, line 90: Should add a stale time, and enabled prop.

Should only be enabled if we have the accessControlList - not having version is fine.

src/hooks/useAntsForWallet.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const { data: antData = {}, isLoading, isRefetching } = useAntsForWallet();
+  const {
+    data: latestAntVersion,
+    isLoading: isLoadingLatestAntVersion,
+    isRefetching: isRefetchingLatestAntVersion,
+  } = useLatestANTVersion();
+  return {
+    ants: Object.keys(antData)
+      .filter((processId) => {
+        return +antData[processId].version < +(latestAntVersion?.version || 0);
```
atticusofsparta, line 106: need to check the role on these ants as well, if its not an Owner role, then needs to be filters out.

Or, if not, then we need to update the permissions/handling on the buttons (which tbh is probably a better idea to show the actual state with a different UI)

PR #829, opened 2025-08-04
fix(PE-8393): do not use class as hash key for use query

atticusofsparta:


src/hooks/useUploadCostGib.ts
```diff
@@ -1,17 +1,26 @@
...
         throw Error('TurboUnauthenticatedClient is not set');
       }
-      return turbo.turboUploader.getUploadCosts({ bytes: [GiB] });
+      const res = await turbo.turboUploader
+        .getUploadCosts({ bytes: [GiB] })
+        .catch((err) => {
+          console.log('err', err);
+          return null;
+        });
+      console.log('res', res);
```
dtfiedler, line 22:
```suggestion
```

dtfiedler (dismissed):

PR #836, opened 2025-08-05
feat(manage): show priority on undernames table on manage page

dtfiedler:


src/components/data-display/tables/DomainsTable.tsx
```diff
@@ -630,7 +630,12 @@ const DomainsTable = ({
...
               }
-              arnsDomain={row.getValue('name')}
+              arnsRecord={{
+                name: row.getValue('name'),
+                version: row.original.version,
+                undernameLimit: row.original.undernames.supported,
+                processId: row.getValue('processId'),
+              }}
               antId={row.getValue('processId')}
               version={row.original.version}
```
atticusofsparta, line 640: If adding the processId and version as the arnsRecord then we can remove antId and version from the other props.

dtfiedler: updated https://github.com/ar-io/arns-react/pull/836/commits/9a8b6470a3885ace823077bdb33d6173407496b1

PR #857, opened 2025-08-13
fix(ant): remove use of all arns records in undername modal

dtfiedler:
## Summary by CodeRabbit

* **New Features**
  * Inline info tooltip added to the undername modal explaining AR.IO length constraints.

* **Performance Improvements**
  * Record fetching moved to paginated loading with longer caching for improved scalability.

* **UX Improvements**
  * Undername modal now accepts a name directly, simplifying the creation flow and removing the previous warning card in favor of the inline tooltip.

* **Tests**
  * Unit tests updated to reflect the new record-fetching approach.


src/components/modals/ant-management/AddUndernameModal/AddUndernameModal.tsx
```diff
@@ -50,13 +48,18 @@ function AddUndernameModal({
   >({});
   const [maxUndernameLength, setMaxUndernameLength] =
     useState<number>(MAX_UNDERNAME_LENGTH);
+  const { data: affiliatedArNSRecords } = useArNSRecords({
+    filters: {
+      processId: [antId],
+    },
+  });
```
atticusofsparta, line 55: I recommend we actually just ignore other domains and use the one in question.

In the args of this component we pass in the ANT id but we can reconfigure to pass in the the domain as well.

Based off that we can predicate the max domain length. Thats the only reason we are querying for records here.

In previous iterations this component suited both manage domain and manage ANT workflows but now that we have just manage domain, needs are different, and I don't fee we get enough value from this query to justify it. Its gonna be rare that theres one ant id connected to multiple names, and even then rare that the name they create will be incompatible on the other one.

What we can do instead to help alert there that is queryless is alert with row colors on the undernames tables for the domains that a undername is too long to be supported on the domain.

TLDR lets do less :)

src/hooks/useArNSRecords.tsx
```diff
@@ -0,0 +1,40 @@
...
+  filters?: Partial<
+    Record<
+      keyof AoArNSNameData,
+      string | number | boolean | string[] | number[] | boolean[]
+    >
+  >;
+} = {}) {
+  const [{ arioContract }] = useGlobalState();
+  return useQuery({
+    queryKey: ['arns-records', arioContract.process.processId, filters],
```
atticusofsparta, line 17: Good hook - however this is where managing the query cache becomes weird.

Right now we have multiple records hooks fetching records. useArNSRecord, and useArNSRecords.

useArNSRecords keys by filters and useArNSRecord keys by the arns name.

.... meaning we may have the data but we will refetch it just because the predicate for the data doesn't match up.

Im going to investigate some patterns we can use for interacting with our api's and caches and follow up in dm's for options.

PR #880, opened 2025-09-22
chore: ArNS Marketplace UI 

dtfiedler:
## Summary by CodeRabbit

* **New Features**
  * Listings area: Active/Completed tabs, search-by-name, pagination, listing details with buy/bid/confirm flows, bids view, settlement/recovery actions, Sell action from My ANTs, multi-step new-listing flow with price-schedule preview, and new My ANs/Listings routes.

* **Style**
  * Global style imports updated to include marketplace UI stylesheet; minor spacing/text-node cleanups.

* **Chores**
  * Added marketplace UI/data and UI libs, date picker, utility libs; build-time DEBUG set to false; new marketplace utilities and query tooling.

### DIFF
--- a/src/components/pages/Listings/ActiveListingsTab.tsx
+++ b/src/components/pages/Listings/ActiveListingsTab.tsx
@@ -2,26 +2,25 @@ import { fetchActiveListings } from '@blockydevs/arns-marketplace-data';
 import {
   ActiveListingTable,
   Card,
-  type Domain,
   Pagination,
   useCursorPagination,
 } from '@blockydevs/arns-marketplace-ui';
+import { useAntsMetadata } from '@src/hooks/listings/useAntsMetadata';
+import { usePrepareListings } from '@src/hooks/listings/usePrepareListings';
 import { useGlobalState } from '@src/state';
 import {
-  BLOCKYDEVS_ACTIVITY_PROCESS_ID,
-  getCurrentListingArioPrice,
+  BLOCKYDEVS_MARKETPLACE_PROCESS_ID,
   marketplaceQueryKeys,
 } from '@src/utils/marketplace';
 import { useQuery } from '@tanstack/react-query';
-import { useNavigate } from 'react-router-dom';
+import { useEffect } from 'react';
 
 const PAGE_SIZE = 20;
 
 const ActiveListingsTab = () => {
-  const navigate = useNavigate();
   const [{ aoClient }] = useGlobalState();
   const pagination = useCursorPagination(PAGE_SIZE);
-
+  const queryAntsMetadata = useAntsMetadata();
   const queryActiveListings = useQuery({
     refetchInterval: 15 * 1000,
     structuralSharing: false,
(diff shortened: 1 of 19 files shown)


src/components/cards/NavMenuCard/NavMenuCard.tsx
```diff
@@ -363,7 +363,16 @@ function NavMenuCard() {
...
+                  </span>
+                  <span className="flex gap-2.5">
+                    <TokenIcon fill="var(--text-grey)" width={16} height={16} />
+                    <Link
+                      to="/my-ants"
+                      className="navbar-link hover gap-2.5 items-center text-white text-sm"
+                    >
+                      My ANTs
+                    </Link>
+                  </span>
```
dtfiedler, line 375: remove and let's add this to the Action Items on the `Manage` table

BartoszSolkaBD: @dtfiedler should every row in Manage table should have the "My ANTs" button that will redirect to My ANTs view?

dtfiedler: i think it makes sense to have a top level navigation item for Marketplace (next to Prices/Returned Names) that navigates to the marketplace, which by default shows your ants, and then an action item for each ant to "Sell this ANT/Name" on each row of the manage table

src/components/pages/Listings/Details/ListingExpiredSection.tsx
```diff
@@ -0,0 +1,91 @@
...
+            },
+            onSuccess: async (data) => {
+              console.log(`cancel success`, { data });
+              void Promise.allSettled([
+                queryClient.refetchQueries({
+                  queryKey: [marketplaceQueryKeys.listings.all],
+                }),
+                queryClient.refetchQueries({
+                  queryKey: [marketplaceQueryKeys.myANTs.all],
+                }),
```
atticusofsparta, line 75: Good pattern - what I would recommend adding to is, is setQueryData.

If the action returns the message result with the patched process state, it can be used to update the client query state, instead of having to wait for the refetch to update the data.

src/components/pages/Listings/ActiveListingsTab.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const navigate = useNavigate();
+  const [{ aoClient }] = useGlobalState();
+  const pagination = useCursorPagination(PAGE_SIZE);
+
+  const queryActiveListings = useQuery({
+    refetchInterval: 15 * 1000,
+    structuralSharing: false,
+    queryKey: marketplaceQueryKeys.listings.list('active', {
+      page: pagination.page,
+      pageSize: pagination.pageSize,
```
atticusofsparta, line 30: should probably have cursor in here as well, since the page number and size may be the same with a different cursor.

src/components/pages/Listings/Confirm.tsx
```diff
@@ -0,0 +1,263 @@
...
+                      console.log(`${operation} success`, { data });
+                      void Promise.allSettled([
+                        queryClient.refetchQueries({
+                          queryKey: [marketplaceQueryKeys.listings.all],
+                        }),
+                        queryClient.refetchQueries({
+                          queryKey: [marketplaceQueryKeys.myANTs.all],
+                        }),
+                      ]);
+                    },
```
atticusofsparta, line 237: see comment regarding reusable mutations - this can be defined on the use[my-mutation] hook and then you just call mutate(params) i believe

src/components/pages/Listings/SearchListingByName.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { useNavigate } from 'react-router-dom';
+
+const SearchListingByName = () => {
+  const [searchValue, setSearchValue] = useState('');
+  const [{ aoClient, arioProcessId }] = useGlobalState();
+  const navigate = useNavigate();
+  const mutationSearch = useMutation({
+    mutationFn: async (name: string) => {
+      return searchANT({
+        name,
```
atticusofsparta, line 21: We need to encode this to ASCII + puny code for searches. Dunno if the data lib  does this internally (good idea to actually if it doesn't already). Likewise when rendering the names we should be encoding to unicode.
```suggestion
        name: lowerCaseDomain(name),
```

atticusofsparta (commented):
general comment - the font seems to have changed in some places (like the profile) and the links are no longer the right color.

I think the marketplace-ui package is leaking tailwind configurations.

src/components/pages/Listings/ActiveListingsTab.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+const PAGE_SIZE = 20;
+
+const ActiveListingsTab = () => {
+  const navigate = useNavigate();
+  const [{ aoClient }] = useGlobalState();
+  const pagination = useCursorPagination(PAGE_SIZE);
+
+  const queryActiveListings = useQuery({
+    refetchInterval: 15 * 1000,
```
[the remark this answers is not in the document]
MateuszPaloszBD: If we want to show users newly added listings in real time, then this interval is necessary. If it’s not that important for us, we can skip it here.

src/components/pages/Listings/Details/Details.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { useQuery } from '@tanstack/react-query';
+import { useParams } from 'react-router-dom';
+
+const Details = () => {
+  const { id } = useParams();
+  const [{ aoClient }] = useGlobalState();
+  const [{ walletAddress }] = useWalletState();
+  const queryDetails = useQuery({
+    enabled: !!id,
+    refetchInterval: 15 * 1000,
```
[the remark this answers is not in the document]
MateuszPaloszBD: For example, this is important for auctions that display bids, because users who are bidding simultaneously need to see those updates. If it’s not something that’s essential, then it can be removed.

src/components/pages/Listings/ActiveListingsTab.tsx
```diff
@@ -0,0 +1,90 @@
...
+    structuralSharing: false,
```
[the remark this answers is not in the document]
MateuszPaloszBD: [https://tkdodo.eu/blog/react-query-render-optimizations#structural-sharing](url)

